PR #2, opened 2024-02-24
Add Task, User services for data manipulation

klaidoshka:
Modified application permissions to allow connecting to the database from within phone.
Added serialization library (very static-typed!)
Added serializations for User/Task models

Introduced a few extension methods for easier work with Json objects. Kotlin's serialization library does not support basic deserialization into Map<String, Any> or serialization into String from any type. So for that, extension methods will be helpful.

It is possible that not all cases will be covered regarding it. But for now, it works.


vytjar (approved):
Good

youreondrugs (approved):

PR #5, opened 2024-02-27
PVP-103 general personal task creation form

AurisTFG:


vytjar (changes requested):
- Should use the Task class located at model -> Tasks.kt and cover all its properties instead of its own data class'
- Task list is probably not needed
- No need to set frequency (I think so, but should confirm with others)

klaidoshka (changes requested):

app/src/main/java/com/pvp/app/ui/screen/task/GeneralTaskCreationForm.kt
```diff
@@ -0,0 +1,247 @@
...
+import androidx.compose.material3.TextButton
+import com.pvp.app.model.Task
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneId
+
+class GeneralTasksManager {
+
+    private val _tasks = mutableStateListOf<Task>()
+    val tasks: List<Task> get() = _tasks
```
klaidoshka, line 42: Redundant abstraction as in the end you are still publicly accessing whole `List` instance and can modify it outside as if it was public. Either use `val tasks = mutableStateListOf<Task>()` or make current `tasks.getter` return `_tasks.toList()` which would create a new immutable `List` instance upon getting every time. If chosen secondary option, you may have to adjust the associated code below.

However, tasks list showcase component (below) is not really necessary as only creation form is needed, which would persist data into the database and return user back to the calendar view where user could see the freshly created task card.

app/src/main/java/com/pvp/app/ui/screen/task/GeneralTaskCreationForm.kt
```diff
@@ -0,0 +1,247 @@
...
+                    isDatePickerDialogOpen.value = false
+                },
+                confirmButton = {
+                    TextButton(
+                        onClick = {
+                            isDatePickerDialogOpen.value = false
+                            val instant = datePickerState.selectedDateMillis?.let {
+                                Instant.ofEpochMilli(it)
+                            }
+                            startDate = LocalDateTime.ofInstant(instant, ZoneId.systemDefault())
```
klaidoshka, line 140: Separate different groups of code by a single line to increase readability.
```kotlin
isDatePickerDialogOpen.value = false // <-- group 1

val instant = datePickerState.selectedDateMillis?.let { // <-- group 2
    Instant.ofEpochMilli(it)
}

startDate = LocalDateTime.ofInstant(instant, ZoneId.systemDefault()) // <-- group 3
```
If you would transfer startDate above variable initialization group (currently group 2), then only 2 groups would exist altogether and only single separation line would be needed between variable assignation and initialization groups.
```kotlin
isDatePickerDialogOpen.value = false // <-- group 1 whole block, doing same action
startDate = LocalDateTime.ofInstant(instant, ZoneId.systemDefault())

val instant = datePickerState.selectedDateMillis?.let { // <-- group 2
    Instant.ofEpochMilli(it)
}
```

app/src/main/java/com/pvp/app/ui/screen/task/GeneralTaskCreationForm.kt
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        Button(
+            onClick = {
+                generalTasksManager.addTask(
+                    title = title,
+                    description = description,
+                    startDate = startDate
+                )
+                title = ""
```
klaidoshka, line 169: A single line is needed after [:168](https://github.com/PVP-K410/Calendar/pull/5/files#diff-5cb4157cc3199595b3f890972f66692335490de0d32dd64c40c87ffd8b102454R168) to separate different blocks of code to increase readability.

klaidoshka (changes requested):

app/src/main/java/com/pvp/app/ui/screen/task/GeneralTaskCreationForm.kt
```diff
@@ -0,0 +1,393 @@
...
+    onCancel: () -> Unit,
+    onConfirm: () -> Unit,
+    toggle: @Composable () -> Unit = {},
+    content: @Composable () -> Unit,
+) {
+    Dialog(
+        onDismissRequest = onCancel,
+        properties = DialogProperties(
+            usePlatformDefaultWidth = false
+        ),
```
klaidoshka, line 340: Unnecessary comma, could be removed.

app/src/main/java/com/pvp/app/ui/screen/task/GeneralTaskCreationForm.kt
```diff
@@ -0,0 +1,393 @@
...
+                horizontalAlignment = Alignment.CenterHorizontally
+            ) {
+                Text(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .padding(bottom = 20.dp),
+                    text = title,
+                    style = MaterialTheme.typography.labelMedium
+                )
+                content()
```
klaidoshka, line 364: Components should be separated by an empty line. Text // content() // Row.

app/src/main/java/com/pvp/app/ui/screen/task/GeneralTaskCreationForm.kt
```diff
@@ -0,0 +1,393 @@
...
+                Row(
+                    modifier = Modifier
+                        .height(40.dp)
+                        .fillMaxWidth()
+                ) {
+                    toggle()
```
klaidoshka, line 370: Components below should be separated by an empty line.

You could also open TextButton lambdas into multiple lines (since everywhere else you chose to do so) or make those components into single line call, since there is only one `onClick` parameter.
```kotlin
TextButton(onClick = onCancel) { Text("Cancel") }
```

app/src/main/java/com/pvp/app/ui/screen/task/GeneralTaskCreationForm.kt
```diff
@@ -0,0 +1,393 @@
...
+                    ) { Text("OK") }
+                }
+            }
+        }
+    }
+}
+
+@Composable
+fun TaskFormAndList() {
+    val generalTasksManager = GeneralTasksManager()
```
klaidoshka, line 386: Could be named `manager` by refactoring (`Shift+F6`). There are no other managers and it would be not complicated, simple, yet easy to understand name.

app/src/main/java/com/pvp/app/ui/screen/task/GeneralTaskCreationForm.kt
```diff
@@ -0,0 +1,393 @@
...
+            duration = Duration.ofMinutes(duration.toLong()),
+            isCompleted = false,
+            scheduledAt = startDate,
+            title = title,
+            userEmail = "TODO"
+        )
+
+        tasks.add(task)
+    }
+
```
klaidoshka, line 82: Redundant empty line.

klaidoshka (approved):
Don't forget to do some extra clean-up.

vytjar (approved):

PR #8, opened 2024-02-29
PVP-106

KhajiitLikeToSneak:


app/src/main/java/com/pvp/app/ui/screen/task/TaskScreen.kt
```diff
@@ -1,7 +1,270 @@
...
+                        descriptionValue.isNotEmpty()
+                    ) {
+                        recipeValue = if (ingredientsValue.isNotEmpty() && preparationValue.isNotEmpty()) {
+                            "$ingredientsValue\n$preparationValue"
+                        } else if (ingredientsValue.isNotEmpty() && preparationValue.isEmpty()) {
+                            ingredientsValue
+                        } else {
+                            preparationValue
+                        }
+                        model.createTaskMeal(
```
klaidoshka, line 238: Different purpose blocks of code should be separated by a single line. One block is assigning value to a variable. Other one is executing some action.

klaidoshka (approved):

PR #15, opened 2024-03-05
PVP-148 Add login process

klaidoshka:
- Added primitive login process that is almost doing the same as register process

In the future validation and inputs form will be added to make these two (login & register) processes different from each other.


vytjar (approved):
Seems good.

PR #18, opened 2024-03-07
PVP 139

vytjar:
Added Health Connect API and a simple screen to view steps.

### DIFF
--- a/app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
+++ b/app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
@@ -2,6 +2,7 @@ package com.pvp.app.ui.screen.steps
 
 import androidx.health.connect.client.records.StepsRecord
 import android.os.Build
+import android.widget.Toast
 import androidx.annotation.RequiresApi
 import androidx.health.connect.client.HealthConnectClient
 import androidx.health.connect.client.PermissionController
@@ -21,8 +22,8 @@ import javax.inject.Inject
 
 @HiltViewModel
 class StepViewModel @Inject constructor(
-    private val healthConnectClient: HealthConnectClient,
-    private val healthConnectService: HealthConnectService
+    private val client: HealthConnectClient,
+    private val service: HealthConnectService
 ) : ViewModel() {
 
     val PERMISSIONS = setOf(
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
```diff
@@ -0,0 +1,65 @@
...
+    suspend fun permissionsGranted(): Boolean {
+        val granted = healthConnectClient.permissionController.getGrantedPermissions()
+        return granted.containsAll(PERMISSIONS)
+    }
+
+    @RequiresApi(Build.VERSION_CODES.UPSIDE_DOWN_CAKE)
+    fun updateTodaysSteps() {
+        viewModelScope.launch {
+            val endTime = Instant.now()
+            val startTime = LocalDate.now().atStartOfDay(ZoneId.systemDefault()).toInstant()
```
klaidoshka, line 47: Names `end` and `start` would be quite understandable too. For as long as there is no other context for `end/start` names. If there were `endDate`, `startDate` together with these times, then `time` and `date` specifiers should stay.

app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
```diff
@@ -0,0 +1,65 @@
...
+
+            try {
+                val stepsRecords: List<StepsRecord> = healthConnectService.readActivityData(
```
klaidoshka, line 50: Name could be simplified to `records`

app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
```diff
@@ -0,0 +1,65 @@
...
+                    recordClass = StepsRecord::class,
+                    startTime = startTime,
+                    endTime = endTime
+                )
+
+                val totalSteps = stepsRecords.sumOf { it.count.toInt() }
+
+                _stepsCount.value = totalSteps
```
klaidoshka, line 58: Above calculation could be put into assign block as a single line.

```kotlin
 _stepsCount.value = stepsRecords.sumOf { it.count.toInt() }
 ```

File: app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt [hunk 4]
```diff
@@ -0,0 +1,65 @@
...
+            } catch (e: Exception) {
+
```
klaidoshka, line 60: Redundant empty line.

In what case an exception could be thrown? If there are multiple possible cases and some of them are because of the user or our used service fault, we should inform user about this error.

If we are going to handle this, a toast could be shown. Toast requires a `Context` object, you could get it in the composable with `LocalContext.current` and then call extended function: `showToast`.

klaidoshka (changes requested):

app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
vytjar: Error should only be thrown if permissions are not sufficient. For example if the composable would not check for permissions first.

As StepViewModel and StepScreen will most likely not be used and deleted later, I don't think it's necessary to fix this.

In case we decide to keep it (or reuse the updateSteps method in other ViewModel), I will add a comment that the exception should be handled.

klaidoshka (approved):

AurisTFG (approved):

PR #19, opened 2024-03-08
PVP-41

KhajiitLikeToSneak:
Added filters functionality
Design could be updated later on

### DIFF
--- a/app/src/main/java/com/pvp/app/ui/screen/filter/Filter.kt
+++ b/app/src/main/java/com/pvp/app/ui/screen/filter/Filter.kt
@@ -26,20 +26,20 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.compose.runtime.LaunchedEffect
+import com.pvp.app.model.SportActivity
 
-@OptIn(ExperimentalLayoutApi::class)
 @Composable
 fun ActivitiesFilter(
     model: FilterViewModel = hiltViewModel()
 ) {
-    val allActivities = listOf("Cycling", "Gym", "Running", "Swimming", "Walking", "Yoga")
+    val activities: List<String> = SportActivity.values().map { it.title }
     var selectedActivities by remember { mutableStateOf(emptyList<String>()) }
     var unselectedActivities by remember { mutableStateOf(emptyList<String>()) }
     val user = model.user.collectAsStateWithLifecycle()
 
     LaunchedEffect(user.value) {
         selectedActivities = user.value?.activities ?: emptyList()
-        unselectedActivities = allActivities - selectedActivities
+        unselectedActivities = activities - selectedActivities
     }
 
     Column(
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/pvp/app/ui/screen/filter/FilterViewModel.kt
```diff
@@ -0,0 +1,38 @@
...
+                .getCurrent()
+                .first()
+        }
+    }
+
+    fun updateUserActivities(activities: List<String>) {
+        viewModelScope.launch {
+            val currentUser = user.value ?: return@launch
+            val updatedUser = currentUser.copy(activities = activities)
+            userService.merge(updatedUser)
```
klaidoshka, line 34: Separation from variables initialization is required.

app/src/main/java/com/pvp/app/ui/screen/filter/FilterViewModel.kt
```diff
@@ -0,0 +1,38 @@
...
+            _user.emit(updatedUser)
```
klaidoshka, line 35: Separation required.

klaidoshka (changes requested):

vytjar (approved):
Functionality seems okay. Will need a pop up or something in the future to inform the user that changes have been applied.

app/src/main/java/com/pvp/app/ui/screen/filter/Filter.kt
```diff
@@ -0,0 +1,132 @@
...
+    ) {
+        Column(
+            modifier = Modifier.fillMaxWidth()
+        ) {
+            Text(
+                text = title,
+                style = TextStyle(fontSize = 18.sp),
+                modifier = Modifier.padding(8.dp)
+            )
+            FlowRow(
```
klaidoshka, line 105: Empty single line could be added to separate components of the same level.

app/src/main/java/com/pvp/app/ui/screen/filter/Filter.kt
```diff
@@ -0,0 +1,132 @@
...
+                                MaterialTheme.shapes.medium
+                            ),
+                        onClick = { onClick(activity) }
+                    ) {
+                        Text(
+                            text = activity,
+                            style = TextStyle(fontSize = 16.sp),
+                            modifier = Modifier.padding(16.dp)
+                        )
+                    }
```
klaidoshka, line 127: Whole card with its children components may not be correctly indented. `Control+Alt+L` would fix this. I think 4 spaces are missing.

app/src/main/java/com/pvp/app/ui/screen/filter/Filter.kt
```diff
@@ -0,0 +1,132 @@
...
+import androidx.compose.material3.Text
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.compose.runtime.LaunchedEffect
+import com.pvp.app.model.SportActivity
+
+@Composable
+fun ActivitiesFilter(
+    model: FilterViewModel = hiltViewModel()
+) {
+    val activities: List<String> = SportActivity.values().map { it.title }
```
klaidoshka, line 35: Could be transfered out of the activities filter component to make it singleton (computed only once) or into `LaunchedEffect` block to make it recompute only after user value changes, but not every time the component is rerendered.

klaidoshka (approved):

PR #22, opened 2024-03-11
PVP-13

KhajiitLikeToSneak:
- Added ingredients filter functionality
- Merged activities and ingredients filter screens so they both use the same functions

### DIFF
--- a/app/src/main/java/com/pvp/app/model/Ingredient.kt
+++ b/app/src/main/java/com/pvp/app/model/Ingredient.kt
@@ -13,23 +13,23 @@ enum class Ingredient(
     val title: String
 ) {
 
-    Chicken("Chicken"),
-    Beef("Beef"),
-    Pork("Pork"),
-    Fish("Fish"),
-    Shrimp("Shrimp"),
     Beans("Beans"),
-    Onions("Onions"),
-    Garlic("Garlic"),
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/pvp/app/ui/screen/filter/Filter.kt
```diff
@@ -132,3 +150,25 @@ fun ActivitiesBox(
...
+
+@Composable
+fun IngredientsFilter(
+    model: FilterViewModel = hiltViewModel()
+) {
+    LaunchedEffect(model) {
+        model.fetchUserData()
+    }
+
+    FilterScreen(title = "ingredients", isActivities = false, model = model)
```
klaidoshka, line 173: Same as above.

klaidoshka (approved):
If that's not too much, you could also refactor `user` model to accept enum lists (SportActivity, Ingredient(?)) and not string lists. I think that would help in later updates. 

Instead of parsing strings into enum values in various methods that use them, we would have them parsed at once in `UserService` by KSerializer.

However, you would also need to handle parsing from strings to enum in the `FilterViewModel.kt`, when merging the user. But that would be the only place.

vytjar (approved):

PR #26, opened 2024-03-12
Pvp 92

vytjar:
Added a monthly calendar view.

- Days that have no tasks are grayed out and unclickable
- When pressed on a day that has tasks that days' view dialog opens


AurisTFG (approved):

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarViewModel.kt
```diff
@@ -1,5 +1,6 @@
 package com.pvp.app.ui.screen.calendar
 
+import android.util.Log
```
klaidoshka, line 3: Is this used, if only import appeared, but not the actual usage of it?

If not used, should remove it.

Also, there is an option to click onto `app` package and select `Optimize Imports` (before the final commit) to clean unused imports through out all the files under the package. `Control+Alt+O` does the same in current file, if I recall correctly.

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarScreen.kt
```diff
@@ -0,0 +1,240 @@
...
+import com.pvp.app.ui.screen.calendar.MonthlyCalendarViewModel
+import com.pvp.app.ui.screen.calendar.getDisplayName
+import java.time.LocalDate
+import java.time.YearMonth
+
+@Composable
+fun MonthlyCalendarScreen(
+    model: MonthlyCalendarViewModel = hiltViewModel(),
+) {
+    val uiState by model.uiState.collectAsState()
```
klaidoshka, line 43: Could be `state`.

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarScreen.kt
```diff
@@ -0,0 +1,240 @@
...
+fun DayDialog(
+    date: LocalDate,
+    tasks: List<Task> = listOf(),
+    onDismissRequest: () -> Unit
+) {
+    Dialog(
+        onDismissRequest = { onDismissRequest() },
+        properties = DialogProperties(usePlatformDefaultWidth = false)
+    ) {
+        Day(name = date.toString(), tasks = tasks)
```
klaidoshka, line 78: Multiple parameters could be split.

app/src/main/java/com/pvp/app/ui/screen/calendar/DateUtil.kt
```diff
@@ -0,0 +1,37 @@
...
+
+object DateUtil {
+
+    val daysOfWeek: Array<String>
+        get() {
+            val daysOfWeek = Array(7) { "" }
+
+            for (dayOfWeek in DayOfWeek.entries) {
+                val localizedDayName =
+                    dayOfWeek.getDisplayName(TextStyle.SHORT, Locale.getDefault())
```
klaidoshka, line 17: Multiple parameters could be split.

Also, `localizedDayName` is resolved quite simply, I believe disposing the single-referenced variable altogether would still keep the code quite readable and _somewhat_ of a problem regarding redundant variable initialization would be solved. At least, I think that it is redundant here.
```kotlin
daysOfWeek[dayOfWeek.value - 1] = dayOfWeek.getDisplayName(
    TextStyle.SHORT,
    Locale.getDefault()
)
```

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarScreen.kt
```diff
@@ -0,0 +1,240 @@
...
+) {
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp)
+    ) {
+        Row {
+            repeat(days.size) {
+                val item = days[it]
+                DayItem(item, modifier = Modifier.weight(1f))
```
klaidoshka, line 97: Multiple parameters could be split.
Single empty line required above.

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarScreen.kt
```diff
@@ -0,0 +1,240 @@
...
+fun Content(
+    dates: List<CalendarUiState.DateEntry>,
+    onClickListener: (CalendarUiState.DateEntry) -> Unit,
+) {
+    Column {
+        var index = 0
+
+        repeat(6) {
+            if (index >= dates.size) {
+                Log.e("CONTENT DATE FILTER", dates.size.toString())
```
klaidoshka, line 171: Single empty line required before return statement.

What is the reason for this log? Could it be more informative (in the logged message or tag)?

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarScreen.kt
```diff
@@ -0,0 +1,240 @@
...
+                return@repeat
+            }
+
+            Row {
+                repeat(7) {
+                    val item =
+                        if (index < dates.size) {
```
klaidoshka, line 178: If possible, make variable initialization condition begin in the same line of where `=` is. If it doesn't fit,  you could try to do something like this:
```kotlin
val item = if (
    index < dates.size
) {
    dates[index]
} else {
    CalendarUiState.DateEntry.Empty
}
```

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarScreen.kt
```diff
@@ -0,0 +1,240 @@
...
+                    index++
+                }
+            }
+        }
+    }
+}
+
+@Composable
+fun ContentItem(
+    date: CalendarUiState.DateEntry,
```
klaidoshka, line 199: Could we name it `entry` instead? To still make it quite understandable to some extent and evade chained `date.date.<...>` calls that by name should return date but first one doesn't.

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarViewModel.kt
```diff
@@ -0,0 +1,114 @@
...
+import java.time.LocalDate
+import java.time.YearMonth
+import javax.inject.Inject
+
+@HiltViewModel
+class MonthlyCalendarViewModel @Inject constructor(
+    private val taskService: TaskService,
+    private val userService: UserService
+) : ViewModel() {
+    private val _currentYearMonth = MutableStateFlow(YearMonth.now())
```
klaidoshka, line 28: Single empty line after class header could be added. (Line :28 would be it)

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarViewModel.kt
```diff
@@ -0,0 +1,114 @@
...
+    private val _uiState = MutableStateFlow(CalendarUiState.Init)
```
klaidoshka, line 29: Could we name it `_state`? Below variable would also become `state`.

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarViewModel.kt
```diff
@@ -0,0 +1,114 @@
...
+
+            combine(flowUser, flowTasks) { user, tasks ->
+                _uiState.update { currentState ->
+                    currentState.copy(
+                        dates = getDates(currentState.yearMonth, tasks),
+                        tasks = tasks
+                    )
+                }
+            }.launchIn(viewModelScope)
+
```
klaidoshka, line 50: Redundant empty line.

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarViewModel.kt
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    fun changeMonth(nextMonth: YearMonth) {
+        _uiState.update {currentState ->
```
klaidoshka: Missing single space. Use Control+Alt+L for the whole file, it will fix all spacing problems for you.

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarViewModel.kt
```diff
@@ -0,0 +1,114 @@
...
+            currentState.copy(
+                yearMonth = nextMonth,
+                dates = getDates(nextMonth, currentState.tasks)
```
klaidoshka, line 58: Multiple parameters could be split into separate lines.

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarViewModel.kt
```diff
@@ -0,0 +1,114 @@
...
+            )
+        }
+    }
+
+    private fun getDates(yearMonth: YearMonth, tasks: List<Task>): List<CalendarUiState.DateEntry> {
+        return yearMonth.getDayOfMonthStartingFromMonday()
```
klaidoshka, line 64: Chained method calls should be split into separate lines.
```kotlin
yearMonth
    .getDayOfMonthStartingFromMonday() // can naming for this method be improved?
    .map {
        // ...
    }
```

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarViewModel.kt
```diff
@@ -0,0 +1,114 @@
...
+    fun changeMonth(nextMonth: YearMonth) {
+        _uiState.update {currentState ->
+            currentState.copy(
+                yearMonth = nextMonth,
+                dates = getDates(nextMonth, currentState.tasks)
+            )
+        }
+    }
+
+    private fun getDates(yearMonth: YearMonth, tasks: List<Task>): List<CalendarUiState.DateEntry> {
```
klaidoshka, line 63: Multiple parameters could be split into separate lines.

app/src/main/res/values/strings.xml
```diff
@@ -26,4 +26,5 @@
     <string name="screen_sign_in_toast_success">Welcome back</string>
     <string name="screen_sign_up_toast_error">Error has occurred while signing up. Try again</string>
     <string name="screen_sign_up_toast_success">Welcome! We are happy to have you here</string>
+    <string name="route_calendar_monthly">Monthly calendar</string>
```
klaidoshka, line 29: Could be inserted after 18 line instead to maintain the order.

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarScreen.kt
```diff
@@ -0,0 +1,240 @@
...
+            Icon(
+                imageVector = Icons.AutoMirrored.Filled.KeyboardArrowRight,
+                contentDescription = "Forward"
+            )
+        }
+    }
+}
+
+@Composable
+fun DayItem(day: String, modifier: Modifier = Modifier) {
```
klaidoshka, line 148: Multiple parameters could be split.

klaidoshka (changes requested):

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarScreen.kt
```diff
@@ -0,0 +1,247 @@
...
+    onClickListener: (CalendarUiState.DateEntry) -> Unit
+) {
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp)
+    ) {
+        Row {
+            repeat(days.size) {
+                val item = days[it]
```
klaidoshka, line 98: You could leave it as it is, but still, writing some ideas for this as I noticed this happens a lot. This is just my opinion.

This is unnecessary and could be called within parameter assignation `day = days[it]`.

Now variable is created O(n) times and only to be referenced in a single spot. This also doesn't change readability for the better, since a very simple action is being made. In a very insignificant way it actually makes the readability worse, more stuff happens by assigning the resolved value.

If the action was longer (like 100+ characters long) or complex (made of multiple steps), then it would make sense to separate these values.

klaidoshka (approved):

PR #30, opened 2024-03-13
Pvp 129

vytjar:
Applied the daily view task design to look like in the picture below

Screenshot_20240313_203423


app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarScreen.kt
```diff
@@ -72,12 +72,26 @@ fun DayDialog(
...
-            tasks = tasks
+            "${date.month.toString().capitalizeFirstLetter()} ${date.dayOfMonth}",
+            tasks = tasks,
+            date = date,
+            expandedUponCreation = true
         )
     }
 }
 
+fun String.capitalizeFirstLetter(): String {
```
klaidoshka, line 83: Pretty sure methods (extensions) for this already exist. Also, put such util elsewhere.

- If utility is required to execute something very specific to associated screen from multiple places, put it under that screen's package.
- If utility can be used in various situations (like now, where it is not tied to some context, it just makes String (used everywhere) to start by a capital latter), it should be placed within `app/common/Utils.kt`.
- If utility can be used in various situations, but only for UI, it should be placed within `/app/ui/common/Utils.kt`

Actual file name can be different, if grouped by some logic.

app/src/main/java/com/pvp/app/ui/screen/calendar/MonthlyCalendarScreen.kt
```diff
@@ -72,12 +72,26 @@ fun DayDialog(
...
+    if (this.isEmpty()) return this
```
klaidoshka, line 84: Should have a body. Should also have a single empty line before return.

app/src/main/java/com/pvp/app/ui/screen/task/TaskScreen.kt
```diff
@@ -548,12 +548,14 @@ fun TaskBox(
...
         modifier = Modifier
             .fillMaxWidth()
+            .padding(8.dp)
             .border(
                 BorderStroke(
                     1.dp,
                     MaterialTheme.colorScheme.outline
                 )
             )
+
```
klaidoshka, line 558: Redundant empty line.

klaidoshka (approved):

AurisTFG (approved):

PR #31, opened 2024-03-14
PVP-141

vytjar:
Added a step counter to the daily view.

- Step counter is displayed only if the Day composable is supplied by the date
- The step progress is displayed in a circle. The step goal is currently hardcoded as 10000. Later we might need to add a way to set a step goal?
- Added a new method inside HealthConnectService for aggregating steps (the old method was prone to duplicates inside the data, although I left it as it allows to read other types of data besides steps)
- Daily view box size was increased so that the step progress circle fits

This is how the step counter looks in dark mode:

Screenshot_20240314_161726

### DIFF
--- a/app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
+++ b/app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
@@ -1,5 +1,8 @@
 package com.pvp.app.ui.screen.calendar
 
+import android.os.Build
+import androidx.activity.compose.rememberLauncherForActivityResult
+import androidx.annotation.RequiresApi
 import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.background
 import androidx.compose.foundation.border
@@ -22,12 +25,15 @@ import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.automirrored.outlined.DirectionsWalk
+import androidx.compose.material.icons.outlined.DirectionsWalk
 import androidx.compose.material.icons.outlined.Place
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
@@ -40,13 +46,16 @@ import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
 import androidx.compose.ui.window.Dialog
+import androidx.health.connect.client.PermissionController
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import com.pvp.app.model.MealTask
 import com.pvp.app.model.SportTask
 import com.pvp.app.model.Task
 import com.pvp.app.ui.common.Button
+import com.pvp.app.ui.screen.steps.StepViewModel
 import com.pvp.app.ui.screen.task.CreateTaskGeneralForm
 import com.pvp.app.ui.screen.task.CreateTaskMealForm
 import com.pvp.app.ui.screen.task.CreateTaskSportForm
@@ -199,7 +208,7 @@ fun Day(
                     shape = RoundedCornerShape(10.dp)
                 )
                 .clickable { expand = !expand }
-                .align(Alignment.CenterHorizontally),
+                .align(Alignment.CenterHorizontally)
         ) {
             Column(
                 modifier = Modifier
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
```diff
@@ -209,41 +218,33 @@ fun Day(
...
-                Text(
-                    "Steps",
-                    textAlign = TextAlign.Center,
-                    modifier = Modifier
-                        .fillMaxWidth()
-                        .padding(16.dp)
-                )
-
-                // Don't display date if no value was supplied
+                // Don't display steps if no date was supplied
```
klaidoshka, line 235: Comment seems to be redundant. I think it is quite obvious from the code.

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
```diff
@@ -291,6 +292,80 @@ enum class TaskFilter(val displayName: String) {
...
+        } else {
+            launcher.launch(PERMISSIONS)
+        }
+    }
+
+    val steps = model.stepsCount.collectAsStateWithLifecycle()
+    val goal = 10000f // TODO create way for user to set a step goal?
+    val progress = steps.value / goal
+
+
```
klaidoshka, line 321: Redundant single line.

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
```diff
@@ -291,6 +292,80 @@ enum class TaskFilter(val displayName: String) {
...
+    Box(
+        contentAlignment = Alignment.Center,
+        modifier = Modifier.fillMaxSize()
+    ) {
+        // These have to be set here because for some reason MaterialTheme.colorScheme
+        // can't be used inside a Canvas ¯\_(ツ)_/¯
```
klaidoshka, line 327: Comment seems redundant.

If I recall correctly, `colorScheme` is a composable. Compiler should show that exact problem - that a composable must be called within other composable function, which Canvas is not, if I am not mistaken.

Either way, an error should be shown by the compiler, if developer would not understand why this is moved to outer space and would move it in. So, I believe, this is unnecessary.

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
```diff
@@ -291,6 +292,80 @@ enum class TaskFilter(val displayName: String) {
...
+                color = progressArcColor,
+                startAngle = -90f,
+                sweepAngle = 360f * progress,
+                useCenter = false,
+                topLeft = topLeft,
+                size = size,
+                style = Stroke(width = strokeWidth, cap = StrokeCap.Round)
+            )
+        }
+        Text(
```
klaidoshka, line 360: Single empty line required.

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarViewModel.kt
```diff
@@ -16,20 +21,28 @@ import kotlinx.coroutines.flow.flowOf
...
 
 @HiltViewModel
 @OptIn(ExperimentalCoroutinesApi::class)
 class CalendarViewModel @Inject constructor(
     private val taskService: TaskService,
-    private val userService: UserService
+    private val userService: UserService,
+    private val healtConnectService: HealthConnectService,
+    private val client: HealthConnectClient
+
```
klaidoshka, line 37: Redundant empty line.

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarViewModel.kt
```diff
@@ -60,10 +73,30 @@ class CalendarViewModel @Inject constructor(
...
+    suspend fun permissionsGranted(): Boolean {
+        val granted = client.permissionController.getGrantedPermissions()
+
+        return granted.containsAll(PERMISSIONS)
+    }
+
+    fun getDaysSteps(date: LocalDate) {
+        viewModelScope.launch {
+            val end = date.plusDays(1).atStartOfDay(ZoneId.systemDefault()).toInstant()
+            val start = date.atStartOfDay(ZoneId.systemDefault()).toInstant()
```
klaidoshka, line 86: Chained methods should be split.

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarViewModel.kt
```diff
@@ -60,10 +73,30 @@ class CalendarViewModel @Inject constructor(
...
+    }
 }
 
 data class CalendarState(
     val tasksMonth: List<Task> = listOf(),
     val tasksWeek: List<Task> = listOf(),
     val user: User? = null
+)
+
+val PERMISSIONS = setOf(
```
klaidoshka, line 99: I believe this should be lowercased, lint code suggestions by Android Studio should show it too. If they don't show it, means I am wrong and it should stay as it is.

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
```diff
@@ -291,6 +292,80 @@ enum class TaskFilter(val displayName: String) {
...
+            val size = Size(radius, radius)
+
+            drawArc(
+                color = backgroundArcColor,
+                startAngle = -90f,
+                sweepAngle = 360f,
+                useCenter = false,
+                topLeft = topLeft,
+                size = size,
+                style = Stroke(width = strokeWidth, cap = StrokeCap.Round)
```
klaidoshka, line 347: Multiple parameters could be split: here, below and in HealthConnectorServiceImpl. Be consistent.

klaidoshka (approved):

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
vytjar: Yes, it throws that error, but

image

Doesn't the @Composable mean it's a composable?

klaidoshka: Yes, but you are providing code for lambda context, which is not a @Composable, but a simple DrawScope instance method.

If it was `DrawScope.@Composable() -> Unit`, then it should work.

vytjar: Ah, makes sense. Thanks

PR #32, opened 2024-03-14
PVP-135

KhajiitLikeToSneak:
Applied design


app/src/main/java/com/pvp/app/ui/screen/settings/Settings.kt
```diff
@@ -0,0 +1,99 @@
...
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(8.dp)
+            .background(
+                MaterialTheme.colorScheme.secondary,
+                MaterialTheme.shapes.medium
+            ),
+    ) {
+        Text(
+            text = " " + title,
```
klaidoshka, line 95: Even shorter would be `" $title"`, but it is fine as it is ^^

klaidoshka (approved):

PR #33, opened 2024-03-14
Applied a simple version of the weekly calendar design

youreondrugs:


app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
```diff
@@ -422,21 +428,46 @@ fun FilterBox(
...
-            Day(
-                name = DayOfWeek.of(it).name,
-                tasks = tasks.filter { task -> task.scheduledAt.dayOfWeek.value == it }
-            )
+    val days = (1..7).map { DayOfWeek.of(it).name }
+    val pagerState = rememberPagerState(pageCount = { days.size })
+    val currentPage = pagerState.currentPage
+
+    @Composable
+    fun DayPage(day: String, pageIndex: Int) {
```
klaidoshka, line 442: Would be easier to read if these components would be exported out of the parent component. If these components are not supposed to be called from outside, you should also make them `private`.

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
```diff
@@ -422,21 +428,46 @@ fun FilterBox(
...
         }
     }
-}
\ No newline at end of file
+
+
+    @Composable
+    fun Page(pageIndex: Int) {
+        val day = days[pageIndex]
+        DayPage(day, pageIndex)
```
klaidoshka, line 463: Single line required between variable initialization and component call.

Also, could just be removed if all it does is just wrap another component. Then, in HorizontalPager iteration just call like this:
```kotlin
DayPage(days[page], page)
```

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
```diff
@@ -422,21 +428,46 @@ fun FilterBox(
...
+    }
+
+    HorizontalPager(
+        state = pagerState,
+        modifier = modifier.padding(horizontal = 16.dp),
+        contentPadding = PaddingValues(start = 32.dp, end = 32.dp),
```
klaidoshka, line 469: Multiple parameters could be split.

klaidoshka (approved):

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarScreen.kt
```diff
@@ -422,21 +428,46 @@ fun FilterBox(
...
+        val tasksForDay = tasks.filter { task -> task.scheduledAt.dayOfWeek.name == day }
+        val scale = animateFloatAsState(
+            targetValue = if (pageIndex == currentPage) 1f else 0.8f,
+            animationSpec = spring(stiffness = 500f)
+        ).value
+
+        Box(
+            modifier = Modifier
+                .fillMaxWidth()
+                .graphicsLayer(scaleX = scale, scaleY = scale),
```
klaidoshka, line 452: Multiple parameters could be split.

KhajiitLikeToSneak (approved):

PR #34, opened 2024-03-14
PVP-90

youreondrugs:


app/build.gradle.kts
```diff
@@ -153,6 +154,9 @@ dependencies {
     androidTestImplementation("androidx.compose.ui:ui-test-junit4")
     debugImplementation("androidx.compose.ui:ui-tooling")
     debugImplementation("androidx.compose.ui:ui-test-manifest")
+
```
klaidoshka, line 157: If possible, put implementation dependencies next to other implementations and not together with test/debug dependencies.

Also, accompanist-permissions is made for permissions 'cleaner' handling, if I am not mistaken. It is not really specific to notifications, could put it just under android group somewhere above.

app/src/main/java/com/pvp/app/Application.kt
```diff
@@ -1,7 +1,20 @@
...
 import android.app.Application
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.content.Context
 import dagger.hilt.android.HiltAndroidApp
 
 @HiltAndroidApp
-class Application : Application()
\ No newline at end of file
+class Application : Application() {
```
klaidoshka, line 10: Single empty line required

app/src/main/java/com/pvp/app/Application.kt
```diff
@@ -1,7 +1,20 @@
...
+    companion object {
+        lateinit var appContext: Context
+            private set
+    }
+
+    override fun onCreate() {
+        super.onCreate()
+        appContext = applicationContext
```
klaidoshka, line 18: Single empty line required.

app/src/main/java/com/pvp/app/model/Tasks.kt
```diff
@@ -86,4 +91,19 @@ open class Task(
     override fun toString(): String {
         return "Task(description=$description, duration=$duration, id=$id, isCompleted=$isCompleted, scheduledAt=$scheduledAt, title='$title', userEmail='$userEmail')"
     }
+
+    fun scheduleReminder(context: Context, minutesToReminder: Int) {
```
klaidoshka, line 95: Multiple parameters could be split. If you are going to follow it, follow throughout the whole function.

app/src/main/java/com/pvp/app/service/NotificationReceiver.kt
```diff
@@ -0,0 +1,76 @@
...
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.os.Build
+import androidx.core.app.ActivityCompat
+import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationManagerCompat
+import com.pvp.app.R
+
+class NotificationReceiver : BroadcastReceiver() {
+    override fun onReceive(context: Context, intent: Intent) {
```
klaidoshka, line 19: Single empty line required.
Parameters could be split, follow throughout the whole file.

app/src/main/java/com/pvp/app/service/NotificationReceiver.kt
```diff
@@ -0,0 +1,76 @@
...
+        showNotification(
+            context,
+            intent.getStringExtra("notificationText"),
+            intent.getIntExtra("notificationId", 0)
+        )
+    }
+
+    private fun showNotification(context: Context, text: String?, notificationId: Int) {
+        val notificationManager = NotificationManagerCompat.from(context)
+        val channelId = "Task"
```
klaidoshka, line 29: If this is channel id and it must be the same for all tasks, I would suggest using `Configuration` interface for this to have a centralized place for constants/configurations. Define something like `val channelNotificationTasksReminderId`.

app/src/main/java/com/pvp/app/service/NotificationReceiver.kt
```diff
@@ -0,0 +1,76 @@
...
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val channel = NotificationChannel(
+                channelId,
+                "Task",
+                NotificationManager.IMPORTANCE_DEFAULT
+            )
+            notificationManager.createNotificationChannel(channel)
```
klaidoshka, line 37: Single empty line required.

app/src/main/java/com/pvp/app/service/NotificationReceiver.kt
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        val notification = NotificationCompat.Builder(context, channelId)
+            .setContentTitle("Calendar reminder")
+            .setContentText(text)
+            .setSmallIcon(R.drawable.ic_launcher_foreground)
+            .setPriority(NotificationCompat.PRIORITY_DEFAULT)
+            .build()
+
+        if (ActivityCompat.checkSelfPermission(
```
klaidoshka, line 47:
```kotlin
        if (
            ActivityCompat.checkSelfPermission(
                context,
                Manifest.permission.POST_NOTIFICATIONS
            ) != PackageManager.PERMISSION_GRANTED
        ) {
            return
        }
```

app/src/main/java/com/pvp/app/service/NotificationReceiver.kt
```diff
@@ -0,0 +1,76 @@
...
+        ) {
+            return
+        }
+
+        notificationManager.notify(notificationId, notification)
+    }
+}
+
+fun scheduleNotification(context: Context, text: String, delaySeconds: Int, notificationId: Int) {
+    println("scheduled after ${delaySeconds}, notification_id = $notificationId")
```
klaidoshka, line 60: Prints used for debugging should not be kept. But if it is, then separate variables from method call by a single empty line.

app/src/main/java/com/pvp/app/service/NotificationReceiver.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    val pendingIntent = PendingIntent.getBroadcast(
+        context,
+        notificationId,
+        intent,
+        PendingIntent.FLAG_IMMUTABLE or PendingIntent.FLAG_UPDATE_CURRENT
+    )
+
+    val triggerAtMillis = System.currentTimeMillis() + (delaySeconds * 1000)
+    alarmManager.setExact(AlarmManager.RTC_WAKEUP, triggerAtMillis, pendingIntent)
```
klaidoshka, line 75: Single empty line required before method call.

klaidoshka (changes requested):

app/src/main/java/com/pvp/app/model/Tasks.kt
```diff
@@ -86,4 +91,30 @@ open class Task(
     override fun toString(): String {
         return "Task(description=$description, duration=$duration, id=$id, isCompleted=$isCompleted, scheduledAt=$scheduledAt, title='$title', userEmail='$userEmail')"
     }
+
+    fun scheduleReminder(
+        context: Context,
+        minutesToReminder: Int)
```
klaidoshka, line 97: Bracket should be in a separate line. `Control+Alt+L` may fix such things.

app/src/main/java/com/pvp/app/service/NotificationReceiver.kt
```diff
@@ -0,0 +1,114 @@
...
+    val intent = Intent(
+        context,
+        NotificationReceiver::class.java
+    )
+
+    intent.putExtra(
+        "notificationText",
+        text
+    )
+    intent.putExtra(
```
klaidoshka, line 95: Because they are of multiple lines, they could be separated by a single line. The same goes for all the multiple-line variables here, above and below too. But on the other hand... it is readable enough. But being inconsistent is bad and so, you _could_ change it. Your choice.

klaidoshka (approved):

AurisTFG (approved):

PR #39, opened 2024-03-15
Notification Hotfix

AurisTFG:
- Fixed notifications crash

### DIFF
--- a/app/src/main/java/com/pvp/app/api/Configuration.kt
+++ b/app/src/main/java/com/pvp/app/api/Configuration.kt
@@ -3,4 +3,5 @@ package com.pvp.app.api
 interface Configuration {
 
     val googleOAuthClientId: String
+    val channelNotificationTasksReminderId: String
 }
--- a/app/src/main/java/com/pvp/app/api/Configuration.kt
+++ b/app/src/main/java/com/pvp/app/api/Configuration.kt
@@ -3,5 +3,4 @@ package com.pvp.app.api
 interface Configuration {
 
     val googleOAuthClientId: String
-    val channelNotificationTasksReminderId: String
 }
--- a/app/src/main/java/com/pvp/app/ui/common/Components.kt
+++ b/app/src/main/java/com/pvp/app/ui/common/Components.kt
@@ -132,7 +132,10 @@ fun ProgressIndicator() {
         modifier = Modifier.fillMaxSize(),
         verticalArrangement = Arrangement.Center
     ) {
-        CircularProgressIndicator()
+        CircularProgressIndicator(
+            color = MaterialTheme.colorScheme.surface,
+            modifier = Modifier.fillMaxWidth(0.5f)
+        )
     }
 }
 
--- a/app/src/main/java/com/pvp/app/ui/common/Components.kt
+++ b/app/src/main/java/com/pvp/app/ui/common/Components.kt
@@ -4,19 +4,27 @@ import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.background
 import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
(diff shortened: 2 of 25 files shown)


app/src/main/java/com/pvp/app/api/Configuration.kt
```diff
@@ -3,5 +3,4 @@ package com.pvp.app.api
 interface Configuration {
 
     val googleOAuthClientId: String
-    val channelNotificationTasksReminderId: String
```
klaidoshka, line 6: Why? This is a configuration, hard-coded into code for now, but still, a configuration to define a name of task notifications.

To fix the NotificationReceiver error you just had to put `@Inject` annotation next to specified primary constructor `class NotificationReceiver @Inject constructor(...)`, since Dagger didn't know where (and how, because of it) to inject.

app/src/main/java/com/pvp/app/ui/common/Components.kt
```diff
@@ -208,4 +220,204 @@ fun TextField(
...
+    onConfirmClick: () -> Unit = {},
+    onDismiss: () -> Unit = {}
+) {
+    var showDialog by remember { mutableStateOf(false) }
+
+    Box(
+        modifier = modifier,
+        contentAlignment = contentAlignment
+    ) {
+        androidx.compose.material3.Button(
```
klaidoshka, line 243: Could use our own button (which is above) to keep being consistent with some of the modifications in the button.

For now our button pretty much changes the shape and possibly something else (I do not recall), in the future we may change it even more. To keep writing those modifications in every spot is a non-sense, so that's why we should create a modified component and re-use it.

app/src/main/java/com/pvp/app/ui/common/Components.kt
```diff
@@ -208,4 +220,204 @@ fun TextField(
...
+            onDismissRequest = {
+                onDismiss()
+
+                showDialog = false
+            },
+            confirmButton = {
+                Box(
+                    contentAlignment = Alignment.BottomEnd
+                ) {
+                    androidx.compose.material3.Button(
```
klaidoshka, line 264: Same as above

app/src/main/java/com/pvp/app/ui/common/Components.kt
```diff
@@ -208,4 +220,204 @@ fun TextField(
...
+                }
+            }
+        )
+    }
+}
+
+@Composable
+fun IconButtonWithDialog(
+    modifier: Modifier = Modifier,
+    icon: ImageVector = Icons.Outlined.Edit,
```
klaidoshka, line 295: Seems weird to have a default icon which is for `edit` purpose in a component that must use icon. Find a better replacement or just make it a required parameter. Don't forget to modify any usage of this component so that no side-effects appear.

app/src/main/java/com/pvp/app/ui/common/Components.kt
```diff
@@ -208,4 +220,204 @@ fun TextField(
...
+    ) {
+        IconButton(
+            onClick = {
+                showDialog = true
+            },
+            modifier = Modifier.size(iconSize)
+        ) {
+            Icon(
+                imageVector = icon,
+                contentDescription = null
```
klaidoshka, line 317: This should be provided for accessibility reasons. You should add an optional parameter for this, which could in default be equal to `null`.

app/src/main/java/com/pvp/app/ui/common/Components.kt
```diff
@@ -208,4 +220,204 @@ fun TextField(
...
+                        }
+                    )
+                }
+            }
+        )
+    }
+}
+
+@Composable
+fun UserInfoItem(
```
klaidoshka, line 365: What is `User` within this component exactly? And what is the purpose of this component? Is it to only showcase information? Then `InfoItem` should be sufficient. But if this also supports editing the value, name must be changed to `InfoEditableItem` or if you are more comfortable with other naming system - `EditableInfoItem`.

klaidoshka (changes requested):

klaidoshka (approved):

KhajiitLikeToSneak (approved):

PR #40, opened 2024-03-15
PVP-68 Add a setting for notifications reminder time

youreondrugs:


app/src/main/java/com/pvp/app/ui/screen/settings/Settings.kt
```diff
@@ -78,6 +85,35 @@ fun CategoryRow(
...
+
+    Slider(
+        value = reminderTimeMinutes.toFloat(),
+        onValueChange = { reminderTimeMinutes = it.toInt() },
+        valueRange = 1f..60f,
+        steps = 59,
+        modifier = Modifier.padding(vertical = 8.dp)
+    )
+
+    Divider(
```
klaidoshka, line 110: I believe this is deprecated. If I am right, IDE should give warnings about it and you should listen to your IDE and refactor code into suggested better alternatives.

app/src/main/java/com/pvp/app/ui/screen/task/TaskViewModel.kt
```diff
@@ -119,7 +119,7 @@ class TaskViewModel @Inject constructor(
             taskService.merge(task)
         }
 
-        task.scheduleReminder(appContext, 10)
+        task.scheduleReminder(appContext, reminderTimeMinutes)
```
klaidoshka, line 122: These parameters could be split into multiple lines.

app/src/main/java/com/pvp/app/ui/screen/settings/Settings.kt
```diff
@@ -78,6 +85,35 @@ fun CategoryRow(
     )
 }
 
+@Composable
+fun ReminderSetting() {
```
klaidoshka, line 89: Whole setting should follow the design of `SettingCard`, now we have a text on the background which is hard to understand. You could adjust the accepted parameters of that component to fill your needs, but make it as abstract as possible and not orientated around your Reminder setting.

app/src/main/java/com/pvp/app/ui/screen/settings/Settings.kt
```diff
@@ -78,6 +85,35 @@ fun CategoryRow(
...
+    Row {
+        Text(
+            text = "Reminder time",
+            style = MaterialTheme.typography.titleLarge,
+        )
+    }
+
+    Text(
+        modifier = Modifier.padding(vertical = 8.dp),
+        text = "Selected (minutes): $reminderTimeMinutes",
```
klaidoshka, line 99: Minutes selected, but what does it mean? What does `Reminder time` setting do? Does it remind me something every `X` minutes? Does it remind me something every `X` minutes before/after task? Or is it something else? User cannot known as it is now and won't be able to use it correctly from the first try.

app/src/main/java/com/pvp/app/ui/screen/settings/Settings.kt
```diff
@@ -78,6 +85,35 @@ fun CategoryRow(
...
+    )
+
+    Slider(
```
klaidoshka, line 102: Maybe our own picker object from `../ui/common/Pickers.kt` would be handy here. Do not use `NumberPicker`, it is a redundant abstraction and I am going to dispose it soon.

I think we could and should use pickers in most places, where user is supposed to pick a value from a _range_.

klaidoshka (changes requested):

app/src/main/java/com/pvp/app/ui/screen/settings/Settings.kt
```diff
@@ -14,15 +14,22 @@ import androidx.compose.material.icons.Icons
...
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
+import androidx.compose.runtime.mutableIntStateOf
+import androidx.compose.runtime.setValue
+import com.pvp.app.ui.common.Picker
+import com.pvp.app.ui.common.PickerState.Companion.rememberPickerState
+
+var reminderTimeMinutes by mutableIntStateOf(10)
```
klaidoshka, line 32: This is implemented really weird, can it be just an int of `10`, because in the end you are just getting the `int`, right?

It also should only be used, if user has not set their value for tasks reminder, else we would take the actual value user has set from before.

Upon setting change the value should be persisted into the database.

app/src/main/java/com/pvp/app/ui/screen/settings/Settings.kt
```diff
@@ -37,18 +44,32 @@ fun SettingsScreen() {
...
 
-        SettingCard(title = "Placeholder")
-        SettingCard(title = "Placeholder")
-        SettingCard(title = "Placeholder")
+        SettingCard(
+            title = "Set Reminder time",
+            value = reminderTimeMinutes,
+            editContent = { value ->
+                val pickerState = rememberPickerState(initialValue = value)
+                Picker(
```
klaidoshka, line 61: Requires a single empty line.

app/src/main/java/com/pvp/app/ui/screen/settings/Settings.kt
```diff
@@ -37,18 +44,32 @@ fun SettingsScreen() {
...
+                    items = (1..60).toList(),
+                    state = pickerState,
+                    startIndex = reminderTimeMinutes - 1,
+                )
+                reminderTimeMinutes = pickerState.value ?: reminderTimeMinutes
```
klaidoshka, line 66: Requires a single empty line.

app/src/main/java/com/pvp/app/ui/screen/settings/Settings.kt
```diff
@@ -95,5 +119,18 @@ fun SettingCard(
             text = " $title",
             fontSize = 18.sp
         )
+
+        if (description != null) {
+            Spacer(modifier = Modifier.size(8.dp))
+            Text(
```
klaidoshka, line 125: Requires a single empty line.

app/src/main/java/com/pvp/app/ui/screen/task/TaskViewModel.kt
```diff
@@ -119,7 +119,10 @@ class TaskViewModel @Inject constructor(
             taskService.merge(task)
         }
 
-        task.scheduleReminder(appContext, 10)
+        task.scheduleReminder(
+            appContext,
+            reminderTimeMinutes
+        )
```
klaidoshka, line 125: Well.. If you changed it here, why did you not change it in other two places? Be consistent. Line 64, 95.

app/src/main/java/com/pvp/app/ui/screen/settings/Settings.kt
```diff
@@ -37,18 +44,32 @@ fun SettingsScreen() {
...
+            editContent = { value ->
+                val pickerState = rememberPickerState(initialValue = value)
+                Picker(
+                    items = (1..60).toList(),
+                    state = pickerState,
+                    startIndex = reminderTimeMinutes - 1,
+                )
+                reminderTimeMinutes = pickerState.value ?: reminderTimeMinutes
+            },
+            description = "Choose how long before the task would you like to be reminded"
```
klaidoshka, line 68: Can user know that it is in minutes?

klaidoshka (changes requested):

klaidoshka:
Refactored notifications and settings handlement into services as we use them within our architecture.
Adjusted SettingCard appearance.

vytjar (approved):

PR #41, opened 2024-03-15
Pvp 102

youreondrugs:


app/src/main/java/com/pvp/app/ui/screen/calendar/VerticalPager.kt
```diff
@@ -0,0 +1,21 @@
...
+import androidx.compose.foundation.pager.rememberPagerState
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+fun VerticalPagerScreen(currentPage: Int) {
+    val pagerState = rememberPagerState(pageCount = { 2 }, initialPage = currentPage)
+
+    VerticalPager(state = pagerState, modifier = Modifier.fillMaxSize()) {page ->
```
klaidoshka, line 15: Multiple parameters should be split. Spacings are not correct, use `Control+Alt+L`.

klaidoshka (changes requested):

app/src/main/java/com/pvp/app/ui/router/Routes.kt
```diff
@@ -78,15 +80,15 @@ sealed class Route(
         iconDescription = "Calendar page button icon",
         path = "calendar",
         resourceTitleId = R.string.route_calendar,
-        screen = { _, _ -> CalendarScreen() }
+        screen = { _, _ -> VerticalPagerScreen(currentPage = 1) }
     )
 
     data object MonthlyCalendar : Route(
```
klaidoshka, line 86: This should be dropped, since Calendar screen is capable of navigating to Monthly calendar and back again. So just `Calendar` route is enough.

I would also suggest renaming `CalendarScreen` file and component into `CalendarWeeklyScreen`, `MonthlyCalendarScreen` into `CalendarMonthlyScreen` and `VerticalPager` into `CalendarScreen`, which would resolve which page currently is being shown.

klaidoshka (approved):

KhajiitLikeToSneak (approved):

PR #42, opened 2024-03-16
PVP-43

KhajiitLikeToSneak:
Added filters survey.
Fixed the issue where the next survey wasn't displayed after completing one.

New bug appeared: after completing the next survey, the earlier survey values in database get reset.
- completed BodyMassIndexSurvey -> all good
- completed FilterActivitiesSurvey -> activities gets updated but mass and height values in database reset to 0
- completed FilterIngredientsSurvey -> ingredients gets updated but activities list reset to emptyList

Couldn't manage to fix the new issue


klaidoshka:
It is because of different coroutine scopes, one doesn't finish executing, but you already updated the state, which is later used to persist the data.

I will try to find alternatives here and if found none, will just use the same scope for these actions with current solution.

klaidoshka:
Using the same solution, can merge when reviews are provided.
Will look for alternatives and if found will pull it via other branch, if this one will be merged.

klaidoshka (approved):

AurisTFG (approved):

klaidoshka (changes requested):
Actually, you forgot to dispose filter routes in navigation drawer. Remove their `resource/strings` too.

klaidoshka (approved):

PR #47, opened 2024-03-18
PVP-166

vytjar:
- Extended the SportActivity enum to contain id's that correspond to HealthConnect ExcersiceSession data type ExerciseTypes
- Added activity reading and display in StepScreen (only for demonstration purposes, if we want to actually display it, it needs to be reworked)

### DIFF
--- a/app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
+++ b/app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
@@ -9,6 +9,7 @@ import androidx.health.connect.client.records.StepsRecord
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.pvp.app.api.HealthConnectService
+import com.pvp.app.model.SportActivity
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
```diff
@@ -40,22 +39,47 @@ class StepViewModel @Inject constructor(
 
     @RequiresApi(Build.VERSION_CODES.UPSIDE_DOWN_CAKE)
     fun updateTodaysSteps() {
+        viewModelScope.launch {
+            val end = LocalDate
+                .now()
+                .plusDays(1)
+                .atStartOfDay(ZoneId.systemDefault())
+                .toInstant()
+            val start = LocalDate
```
klaidoshka, line 48: Variables that are being manipulated in any way (initialized, assigned) with multiple lines block of code should be separated by a single line to increase readability.
```kotlin
            val end = LocalDate
                .now()
                .plusDays(1)
                .atStartOfDay(ZoneId.systemDefault())
                .toInstant()

            val start = LocalDate
                .now()
                .atStartOfDay(ZoneId.systemDefault())
                .toInstant()
```

app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
```diff
@@ -40,22 +39,47 @@ class StepViewModel @Inject constructor(
...
-                    record = StepsRecord::class,
-                    start = start,
-                    end = end
-                )
-
-                _stepsCount.value = records.sumOf { it.count.toInt() }
-            } catch (e: Exception) {
-                // Exception should be handled by showing a Toast
-                e.printStackTrace()
+            val start = LocalDate
```
klaidoshka, line 63: Same as above.

app/src/main/java/com/pvp/app/ui/screen/steps/StepViewModel.kt
```diff
@@ -40,22 +39,47 @@ class StepViewModel @Inject constructor(
...
+                .now()
+                .minusDays(29)
+                .atStartOfDay(ZoneId.systemDefault())
+                .toInstant()
+
+            val records: List<ExerciseSessionRecord> = service.readActivityData(
```
klaidoshka, line 69: Explicit type may be redundant, if readActivityData returns `List<T>`, where `T` is record of `KClass<T>`.

klaidoshka (changes requested):

klaidoshka (approved):

KhajiitLikeToSneak (approved):

PR #49, opened 2024-03-20
REFACTOR PVP-68

klaidoshka:
- Changed the way state is handled
- Fixed task creation bugs regarding user email
- Fixed wrong message being shown in task reminder notification


app/src/main/java/com/pvp/app/ui/screen/layout/LayoutScreenAuthenticated.kt
```diff
@@ -180,6 +181,12 @@ fun LayoutScreenAuthenticated(
                 val stateLayout by viewModel.state.collectAsStateWithLifecycle()
                 val toggleDialog = remember { { isOpen = !isOpen } }
 
+                if (stateLayout.userAvatar == null) {
+                    ProgressIndicator()
+
+                    return@ModalNavigationDrawer
+                }
+
```
klaidoshka, line 189: On first sign-in without any cache would take time to resolve avatar and it would lead into crash as it is used as non-null object below. Hence, this is needed.

youreondrugs (approved):

PR #52, opened 2024-03-21
PVP 172

vytjar:
Added reading of active and total calories burned

### DIFF
--- a/app/src/main/java/com/pvp/app/service/HealthConnectServiceImpl.kt
+++ b/app/src/main/java/com/pvp/app/service/HealthConnectServiceImpl.kt
@@ -33,8 +33,7 @@ class HealthConnectServiceImpl @Inject constructor(
                 )
             )
 
-            response[ActiveCaloriesBurnedRecord.ACTIVE_CALORIES_TOTAL]?.inCalories
-            0.0
+            response[ActiveCaloriesBurnedRecord.ACTIVE_CALORIES_TOTAL]?.inCalories ?: 0.0
         } catch (e: Exception) {
             0.0
         }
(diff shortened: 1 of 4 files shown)


KhajiitLikeToSneak (approved):

app/src/main/java/com/pvp/app/service/HealthConnectServiceImpl.kt
```diff
@@ -48,26 +78,37 @@ class HealthConnectServiceImpl @Inject constructor(
...
             )
 
-            response[DistanceRecord.DISTANCE_TOTAL]?.inMeters ?: 0.0
+            response[TotalCaloriesBurnedRecord.ENERGY_TOTAL]?.inCalories ?: 0.0
         } catch (e: Exception) {
-            Log.e("DISTANCE", "EXCEPTION ${e.message}")
-
             0.0
         }
     }
```
klaidoshka, line 100: Aren't all these methods in the end doing the same, except final property (inCalories, inMeters) is different?
If that's the case, could just make one method and reuse it instead of copy/paste. Most likely of private visibility.

app/src/main/java/com/pvp/app/service/HealthConnectServiceImpl.kt
```diff
@@ -48,26 +78,37 @@ class HealthConnectServiceImpl @Inject constructor(
...
+
+    override suspend fun <T : Record> readActivityData(
+        record: KClass<T>,
+        start: java.time.Instant,
+        end: java.time.Instant
+    ): List<T> {
+        val request = ReadRecordsRequest(
+            recordType = record,
+            timeRangeFilter = TimeRangeFilter.between(start, end)
```
klaidoshka, line 109: Split params.

klaidoshka (approved):

PR #54, opened 2024-03-22
PVP-183 Improve authentication process design

klaidoshka:
# Changes
- Applied improved authentication design
- Refactor minor things to make it look more smooth (mostly because of the consistent 'sunset' background)
- Moved height/mass ranges into Configuration class
- Merged sign-in/sign-up into one authenticate process
- Removed success toasts for sign-out and authenticate
- Adjusted survey screen and forms to look correct within new hard-coded light-theme-like background

# Unauthenticated preview in 3 according states
- Waiting for user input to authenticate
- Authenticating...
- Loading data (looking if user is authenticated upon application restart)

# Surveys preview
- Body Mass Index
- Filters

# Questions
- Should I add a title to the Body Mass Index survey?
- Google button on-click effect is not taking the whole background into account, I believe it is because of wrong order of modifiers being set in the internals. It is a pain to fix... should I try it? Or leave it in the future?


KhajiitLikeToSneak:
> # Questions
> * Should I add a title to the Body Mass Index survey?
> * Google button on-click effect is not taking the whole background into account, I believe it is because of wrong order of modifiers being set in the internals. It is a pain to fix... should I try it? Or leave it in the future?

- Maybe some title like "blah blah in order to calculate your body mass index" or something like that I don't really know. But even without title selecting mass and height options seems straightforward. 
- Leave for later

KhajiitLikeToSneak (approved):

vytjar:
- At least for me, in this context the title does not seem necessary
- Maybe we should leave it for now and consider fixing it in the future. Time would be better spent trying to complete other tasks in time for presentation.

app/src/main/java/com/pvp/app/ui/screen/authentication/AuthenticationScreen.kt
```diff
@@ -0,0 +1,228 @@
...
+    Color("#aa98ff".toColorInt()),
+    Color("#4380ef".toColorInt())
+)
+    .shuffled()
+    .take(3)
+    .map { it.lighten(0.075f) }
+
+@Composable
+private fun Authentication(
+    authenticate: (isOneTap: Boolean, request: Intent) -> Unit,
```
vytjar, line 74: Multiple parameters can be split into multiple lines

app/src/main/java/com/pvp/app/ui/screen/authentication/AuthenticationScreen.kt
```diff
@@ -0,0 +1,228 @@
...
+    ) -> Unit,
+    verticalArrangement: Arrangement.Vertical = Arrangement.Center,
+) {
+    val textBegin = stringResource(R.string.screen_authentication_button_begin)
+
+    val launcher = rememberLauncherForActivityResult(
+        contract = ActivityResultContracts.StartActivityForResult(),
+        onResult = { result ->
+            if (result.resultCode == RESULT_OK) {
+                authenticate(false, result.data ?: return@rememberLauncherForActivityResult)
```
vytjar, line 89: Here too

app/src/main/java/com/pvp/app/ui/screen/authentication/AuthenticationScreen.kt
```diff
@@ -0,0 +1,228 @@
...
+            }
+        }
+    )
+
+    val launcherOneTap = rememberLauncherForActivityResult(
+        contract = ActivityResultContracts.StartIntentSenderForResult(),
+        onResult = { result ->
+            if (result.resultCode == RESULT_OK) {
+                authenticate(true, result.data ?: return@rememberLauncherForActivityResult)
```
vytjar, line 98: Here too

vytjar (approved):

PR #55, opened 2024-03-22
PVP-171 PVP-173 Add points calculation algorithm and points adding functionality

klaidoshka:
# Changes
- Refactored Task model to contain Points model which holds:
  - claimedAt (LocalDateTime)
  - value (Int)
- SportTask changes:
  - `isDaily` moved under this task type
  - `distance` is calculated in kilometers (0.1 would mean 100 meters)
- Points of the task are precalculated upon task creation
- User gets points after completing a task by calling TaskService#claim. Single time.
- Task creation is now left to the service instead of creating them manually
- Implemented PointService in a very basic way

# Question
- Should the algorithm of PointService be further improved? Should I assign some index to SportActivity (where each activity would hold variating value between 0~10) and it would be used to some extent in the points calculation when determining how many points to give for task duration, distance? Since some activities are quite hard to do, so in a way they should give more points.


KhajiitLikeToSneak:
> # Question
> * Should the algorithm of PointService be further improved? Should I assign some index to SportActivity (where each activity would hold variating value between 0~10) and it would be used to some extent in the points calculation when determining how many points to give for task duration, distance? Since some activities are quite hard to do, so in a way they should give more points.

I think it would be more fair and more rewarding for users who complete harder activities. So yes?

klaidoshka:
> > # Question
> > 
> > * Should the algorithm of PointService be further improved? Should I assign some index to SportActivity (where each activity would hold variating value between 0~10) and it would be used to some extent in the points calculation when determining how many points to give for task duration, distance? Since some activities are quite hard to do, so in a way they should give more points.
> 
> I think it would be more fair and more rewarding for users who complete harder activities. So yes?

Will take care of it tomorrow.

vytjar (changes requested):
There is a problem with the forms. Even if the activity should not support distance metrics - the user is allowed to choose the distance. For example boxing for 100km.

Another problem is that when you calculate distance points, even if the ratio is 0 it still credits max points, as the formula (cosh(ln(distance * ratio))) returns infinity when distance * ratio = 0. This might not be a problem, if during task creation we would not allow users to set distance for activities that should not support the distance metric. Or we could skip the calculation if the ratio is 0.

vytjar (approved):

PR #61, opened 2024-03-25
PVP 178

vytjar:
Added suggestions for users to try new activities (currently I named them Weekly Activities):

- New activities for the user to try are generated once a week and presented to the user as a notification 
- These activites are saved as a user property _weeklyActivities_ (later they will be used to increase point yield)

During generation a simple "filtering" regarding disabilities is implemented:

- If we find that the user uses a wheelchair through HealthConnect exercises, we do not suggest user to try new activities, only _Wheelchair_ activity is added to the weeklyActivities
- If a user does not use a wheelchair - we do not suggest him to try the _Wheelchair_ activity

It should be improved and reworked in the next sprint to cover different injuries and disabilities.

Currently the notification text looks like this:

image


app/src/main/java/com/pvp/app/worker/WeeklyActivityWorker.kt
```diff
@@ -0,0 +1,102 @@
...
+                        " and " + activities.last().title +
+                        " will give you more points this week! "
+            }
+        }
+    }
+
+    private suspend fun getRandomInfrequentActivities(count: Int = 4): List<SportActivity> {
+        val activities = exerciseService.getInfrequentActivities().toMutableList()
+
+        return when (activities.contains(SportActivity.Wheelchair)) {
```
klaidoshka, line 83: Is this logic fine? Currently, when user has done Wheelchair activities, we drop it and give other activities (that possibly won't exist), else if user hasn't done Wheelchair activities, we give it to the user. Is it right?

app/src/main/java/com/pvp/app/worker/WeeklyActivityWorkerSetup.kt
```diff
@@ -0,0 +1,39 @@
...
+
+        workManager.enqueueUniquePeriodicWork(
+            WeeklyActivityWorker.WORKER_NAME,
+            ExistingPeriodicWorkPolicy.CANCEL_AND_REENQUEUE,
+            requestPeriodic
+        )
+
+        return Result.success()
+    }
+}
```
klaidoshka, line 39: This _setup_ worker most likely can be extracted out of here into Application's periodic worker start method (your's), since all it does is just create periodic worker a minute after application installation. You probably could do a simple async coroutine delay before starting it (function will have to be suspendable and use `withContext` 🤔)

In my case, I used _setup_ worker because I had to delay the initialization of periodic worker for 0-23:59:59 interval (from least, to most possible time, theoretically). Delay can be assigned only to the OneTimeWorker, that's why another worker was required. Couldn't depend on coroutine delay, because the time is too great, it may not survive in memory, but workers persist in local storage (you could say).

app/src/main/java/com/pvp/app/service/ExerciseServiceImpl.kt
```diff
@@ -28,4 +53,14 @@ class ExerciseServiceImpl @Inject constructor(
...
         )
     }
+
+    override suspend fun getInfrequentActivities(maxOccurrence: Int): List<SportActivity> {
+        val activityOccurrences = getActivitiesWithOccurrencesMap()
+
+        return SportActivity.entries.filter { activity ->
+            (activityOccurrences[activity] == null
+                    || activityOccurrences[activity]!! <= maxOccurrence)
+                    && activity.id != 0
```
klaidoshka, line 63: What does `0` id stand for? It most likely would be better to reference exact SportActivity and then get its id to not hard-code it and make it more readable.

`SportActivity.X.id`

app/src/main/java/com/pvp/app/worker/WeeklyActivityWorker.kt
```diff
@@ -0,0 +1,102 @@
...
+
+                    Result.success()
+                } catch (e: Exception) {
+                    Result.failure()
+                }
+            } ?: Result.failure()
+    }
+
+    private fun formNotificationBody(activities: List<SportActivity>): String {
+        return when (activities.count() == 1) {
```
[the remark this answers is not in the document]
vytjar: It's the first case, references Collection#size.

app/src/main/java/com/pvp/app/service/ExerciseServiceImpl.kt
vytjar: It stands for `Other`, helps handle activities we do not support.

app/src/main/java/com/pvp/app/service/ExerciseServiceImpl.kt
```diff
@@ -3,15 +3,40 @@ package com.pvp.app.service
...
+            .minusDays(29)
+            .atStartOfDay(ZoneId.systemDefault())
+            .toInstant()
+
+        return service
+            .readActivityData(
+                record = ExerciseSessionRecord::class,
+                start = start,
+                end = end
+            ).map { record ->
```
klaidoshka, line 33: Split

app/src/main/java/com/pvp/app/worker/WeeklyActivityWorker.kt
```diff
@@ -0,0 +1,107 @@
...
+
+                    else -> {
+                        "Participating in ${activities.first().title} " +
+                                "will give you more points this week!"
+                    }
+                }
+            }
+
+            else -> {
+                activities.dropLast(1).joinToString(separator = ", ") { it.title } +
```
klaidoshka, line 73: Split

app/src/main/java/com/pvp/app/service/ExerciseServiceImpl.kt
```diff
@@ -28,4 +54,14 @@ class ExerciseServiceImpl @Inject constructor(
             )
         )
     }
+
+    override suspend fun getInfrequentActivities(maxOccurrence: Int): List<SportActivity> {
+        val activityOccurrences = getActivitiesWithOccurrencesMap()
+
+        return SportActivity.entries.filter { activity ->
+            (activityOccurrences[activity] == null
+                    || activityOccurrences[activity]!! <= maxOccurrence)
```
klaidoshka, line 63: All operators in all cases could instead be within the initial line and then split upon combined entry. Just like lambdas.
```kotlin
           (activityOccurrences[activity] == null ||
               activityOccurrences[activity]!! <= maxOccurrence) &&
           activity != SportActivity.Other
// or...
           (
               activityOccurrences[activity] == null ||
               activityOccurrences[activity]!! <= maxOccurrence // this may have to be indented by 4 spaces in kotlin, not sure (Control+Alt+L)
           ) &&
           activity != SportActivity.Other
```

klaidoshka (approved):

KhajiitLikeToSneak (approved):

PR #62, opened 2024-03-26
PVP 179

vytjar:
Increased yield of points when sports activity is users weekly activity (a multiplier is applied when calculating distance/duration points, currently set at 2)

Also added a hotfixes for:

- Crashing when trying to complete a newly created task
- Checkbox state of tasks being saved between days

### DIFF
--- a/app/src/main/java/com/pvp/app/service/PointServiceImpl.kt
+++ b/app/src/main/java/com/pvp/app/service/PointServiceImpl.kt
@@ -26,6 +26,8 @@ class PointServiceImpl @Inject constructor(
     private val userService: UserService
 ) : PointService {
 
+    private val weeklyMultiplier = 2
+
     override suspend fun calculate(
         task: Task,
         increasePointYield: Boolean
@@ -197,6 +199,4 @@ class PointServiceImpl @Inject constructor(
             }
         }
     }
-
-    private val weeklyMultiplier = 2
 }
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/pvp/app/service/PointServiceImpl.kt
```diff
@@ -188,4 +197,6 @@ class PointServiceImpl @Inject constructor(
             }
         }
     }
+
+    private val weeklyMultiplier = 2
```
klaidoshka, line 201: Can we instead put it to the top of the class?

app/src/main/java/com/pvp/app/service/TaskServiceImpl.kt
```diff
@@ -309,6 +312,12 @@ class TaskServiceImpl @Inject constructor(
             }
     }
 
+    private suspend fun isWeekly(activity: SportActivity): Boolean{
```
klaidoshka, line 315: Space missing

app/src/main/java/com/pvp/app/service/TaskServiceImpl.kt
```diff
@@ -309,6 +312,12 @@ class TaskServiceImpl @Inject constructor(
...
+        return userService.user
+            .firstOrNull()?.weeklyActivities
+            ?.contains(activity) ?: false
```
klaidoshka, line 318: I tend to think of the **Elvis** operator as `orElse` method and here would split it into the new line. I guess it is fine like it is. Up to each preferences, but in the end it would be best if we all handled it the same way.

klaidoshka (approved):

KhajiitLikeToSneak (approved):

PR #63, opened 2024-03-26
CHORE Experience showcase within profile

klaidoshka:
# Changes
- Changed how experience is calculated
- Cleaned up profile editing a little
- Added missing experience, points showcase to the profile and applied other things according to the design

# Levels with respective experience total amount
image

### DIFF
--- a/app/src/main/java/com/pvp/app/api/ExperienceService.kt
+++ b/app/src/main/java/com/pvp/app/api/ExperienceService.kt
@@ -14,19 +14,6 @@ interface ExperienceService {
     ): Int
 
     /**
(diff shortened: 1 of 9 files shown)


vytjar:
Screenshot_20240327_014553

Might be bugged since I had experience before this update.

vytjar (approved):

AurisTFG (approved):

klaidoshka:
> Leveling might be bugged since I had experience before this update.
> 
> Should we reset experience and points to zero so it'd work correctly?
> 
> Screenshot_20240327_014553

Should automatically be fixed upon first task claim. Since exp is taken into account on every claim.

PR #64, opened 2024-03-27
PVP-167 Notification permission handler

AurisTFG:
# Changes:
- Implemented notification permission handler:
   - Detect whether notifications are disabled for the entire app or for individual notification channels.
   - Display a dialog upon app launch if notifications are disabled, providing an option to navigate to settings.
- Updated Gradle to version 8.3.1.

### DIFF
--- a/app/src/main/java/com/pvp/app/Activity.kt
+++ b/app/src/main/java/com/pvp/app/Activity.kt
@@ -1,11 +1,18 @@
 package com.pvp.app
 
+import android.app.AlertDialog
+import android.app.NotificationManager
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
 import android.os.Bundle
+import android.provider.Settings
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.material3.Surface
 import androidx.compose.ui.Modifier
+import androidx.core.app.NotificationManagerCompat
 import com.pvp.app.ui.router.Route
 import com.pvp.app.ui.screen.layout.LayoutScreenBootstrap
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/pvp/app/Activity.kt
```diff
@@ -38,4 +49,51 @@ class Activity : ComponentActivity() {
...
+            if (channel.importance == NotificationManager.IMPORTANCE_NONE) {
+                return false
+            }
+        }
+
+        return true
+    }
+
+    private fun openNotificationSettingsForApp(context: Context) {
+        val intent = Intent().apply {
```
klaidoshka, line 77: Split

app/src/main/java/com/pvp/app/Activity.kt
```diff
@@ -38,4 +49,51 @@ class Activity : ComponentActivity() {
...
+            action = Settings.ACTION_APPLICATION_DETAILS_SETTINGS
+            addCategory(Intent.CATEGORY_DEFAULT)
```
klaidoshka, line 79: Separate

app/src/main/java/com/pvp/app/Activity.kt
```diff
@@ -38,4 +49,51 @@ class Activity : ComponentActivity() {
...
+            data = Uri.parse("package:" + context.packageName)
+        }
+
+        context.startActivity(intent)
+    }
+
+    private fun showNotificationPermissionDialog(context: Context) {
+        AlertDialog.Builder(context)
```
klaidoshka, line 87: Can leave as it is, since technically it is a path into inner Builder class's constructor, or can split after AlertDialog. I guess it is better to not split. But writing the idea behind why.

klaidoshka (approved):

vytjar (approved):

PR #68, opened 2024-03-28
PVP-169 & PVP-177

youreondrugs:
PVP-169:

Daily tasks should be generated according to the activities user likes (the processed activity data)

After generation these tasks should be assigned to the user.

PVP-177:

Added daily task difficulty add display of daily tasks in weekly and monthly views.

Difficulty calculation is based on:

- Level of user activity, that is derived using users activity of the last 30 days, taken from Health Connect. Leveling is based on walking distance, for example if a user walks 4 km everyday and doesn't do any other exercises - his activity level would be 4. Ratios present in SportActivity enum for each activity are used to scale other activities, for example running 4km every day would give you a level bigger than 4.
- Level is taken into account in daily task creation and used as a multiplier from base values.


app/src/main/java/com/pvp/app/api/TaskService.kt
```diff
@@ -98,7 +98,7 @@ interface TaskService : DocumentsCollection {
     suspend fun generateDaily(
         count: Int = 3,
         userEmail: String
-    )
+    ): List<SportTask> // changed
```
klaidoshka, line 101: Comment reason?

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarWeeklyScreen.kt
```diff
@@ -408,10 +409,27 @@ fun ActivitiesBox(
...
-                    TodaysTasks(Icons.AutoMirrored.Outlined.LibraryBooks, tasks, Task::class)
-                    TodaysTasks(Icons.AutoMirrored.Outlined.DirectionsRun, tasks, SportTask::class)
-                    TodaysTasks(Icons.Outlined.Restaurant, tasks, MealTask::class)
+                    DayTasks(
+                        Icons.Outlined.Event,
+                        tasks,
+                        SportTask::class,
+                        true
+                    )
+                    DayTasks(
```
klaidoshka, line 418: Separate

klaidoshka (approved):

KhajiitLikeToSneak (approved):

PR #72, opened 2024-04-09
PVP-229 Sort tasks in tasks preview

klaidoshka:
# Very big diff...
However, most of the code only moved to different place. Below I wrote most specific things that changed so that you could look into wanted parts more in-depth.

# Changes
- Separated utility methods into separate files by appropriate group, added comments to some of them.
- Separated calendar components into separate files
  - Fixed analyses in monthly calendar, where switch is available, being not in the center
  - Made weekly calendar content of tasks/analyses take more space, since it is unused
  - Made analyses component a little bigger size
  - Task cards show time in 12 hours format now to keep it consistent with preview/edit forms
    - ..._Or we could change both to 24 hours format._..
- Split Task#scheduledAt into Task#date and Task#time
  - Fixed some of the logic around parts that used Task#scheduledAt. Didn't refactor much to support nullable time (which is possible), so everywhere added a default value of midnight (00.00)
- Fixed a few crashes introduced by some previous commits
  - Also removed redundant `User#dailyTasks`, reintroduced by previous commits that were behind with their HEAD commit while PR'ing.
- Made Task#time changes to always be persisted without seconds, nano-seconds (set to 0) for better sorting support
  - I believe we only want to be specific to minutes at most and not any deeper. So this should be fine.
- Created abstract `PickerPair`, `PickerTime` now uses it too. For quick and simple enough double pickers implementation.
- Added new configuration `rangeKilometers` used for distance
  - It is used within task creation/edit forms. Meters range is not added, because I think that value no matter what must be in 0-999 interval, steps can differ. Currently using 0-900 with step 100.


vytjar (approved):
We should decide between using 12/24 hour format, because currently we use both.

In task cards we display time in 24 hour format with AM/PM indicators, which are quite useless.

image

Whereas in edit forms we used 12 hour format:

image

Personally, I think it'd make most sense to use the 24 hour format, as the 12 hour format is rarely used in Europe.

KhajiitLikeToSneak (approved):

PR #73, opened 2024-04-10
PVP-220

vytjar:
Added task autocomplete using Health Connect records.

Foreground service is necessary as Health Connect does not allow reading of data in background.

### DIFF
--- a/app/src/main/java/com/pvp/app/Application.kt
+++ b/app/src/main/java/com/pvp/app/Application.kt
@@ -15,6 +15,7 @@ import com.pvp.app.model.NotificationChannel
 import com.pvp.app.worker.DailyTaskWorker
 import com.pvp.app.worker.DailyTaskWorkerSetup
 import com.pvp.app.worker.DrinkReminderWorker
+import com.pvp.app.worker.TaskAutocompleteWorker
 import com.pvp.app.worker.TaskPointsDeductionWorkerSetup
 import com.pvp.app.worker.WeeklyActivityWorker
 import dagger.hilt.android.HiltAndroidApp
@@ -48,6 +49,8 @@ class Application : Application(), Configuration.Provider {
         createDrinkReminderWorker()
 
         createNotificationChannels()
+
+        createTaskAutocompleteWorker()
 
         createTaskPointsDeductionWorker()
 
@@ -115,6 +118,22 @@ class Application : Application(), Configuration.Provider {
                 manager.createNotificationChannel(notificationChannel)
             }
         }
+    }
+
+    private fun createTaskAutocompleteWorker(){
+        Log.e("AUTOCOMPLETE", "CREATING")
+        val requestPeriodic = PeriodicWorkRequestBuilder<TaskAutocompleteWorker>(
+            repeatInterval = 15,
+            repeatIntervalTimeUnit = TimeUnit.MINUTES
+        )
+            .build()
+
+        workManager.enqueueUniquePeriodicWork(
+            TaskAutocompleteWorker.WORKER_NAME,
+            ExistingPeriodicWorkPolicy.CANCEL_AND_REENQUEUE,
+            requestPeriodic
+        )
+        Log.e("AUTOCOMPLETE", "CREATED")
     }
 
     private fun createTaskPointsDeductionWorker() {
--- a/app/src/main/java/com/pvp/app/Application.kt
+++ b/app/src/main/java/com/pvp/app/Application.kt
@@ -120,7 +120,7 @@ class Application : Application(), Configuration.Provider {
         }
     }
 
-    private fun createTaskAutocompleteWorker(){
+    private fun createTaskAutocompleteWorker() {
         val requestPeriodic = PeriodicWorkRequestBuilder<TaskAutocompleteWorker>(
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/pvp/app/Application.kt
```diff
@@ -117,6 +120,22 @@ class Application : Application(), Configuration.Provider {
         }
     }
 
+    private fun createTaskAutocompleteWorker(){
+        Log.e("AUTOCOMPLETE", "CREATING")
```
klaidoshka, line 124: Aren't these for debugging while developing? Change to `Log.d` for debug type instead of error type or remove.

app/src/main/java/com/pvp/app/service/TaskAutocompleteService.kt
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    private fun createNotificationChannel() {
+        val channel = NotificationChannel(
+            CHANNEL_ID,
+            "Task Processing Service",
+            NotificationManager.IMPORTANCE_DEFAULT
+        )
+
+        getSystemService(NotificationManager::class.java)?.createNotificationChannel(channel)
```
klaidoshka, line 118: Split

app/src/main/java/com/pvp/app/service/TaskAutocompleteService.kt
```diff
@@ -0,0 +1,192 @@
...
+        return userService.user
+            .firstOrNull()
+            ?.let { user ->
+                taskService
+                    .get(userEmail = user.email)
+                    .map { tasks ->
+                        tasks.filter { task ->
+                            task is SportTask &&
+                                    task.date
+                                        .isEqual(LocalDate.now())
```
klaidoshka, line 142: If it fits, should be in same line as `task.date`

app/src/main/java/com/pvp/app/service/TaskAutocompleteService.kt
```diff
@@ -0,0 +1,192 @@
...
+    @Inject
+    lateinit var taskService: TaskService
+
+    @Inject
+    lateinit var userService: UserService
+
+    companion object {
+
+        private const val NOTIFICATION_ID = Int.MAX_VALUE
+        private const val CHANNEL_ID = "Task Autocomplete"
```
klaidoshka, line 49: I think we are using `NotificationChannel` under `model` package instead. For storing channels with their ids and possibly other data.

app/src/main/java/com/pvp/app/service/TaskAutocompleteService.kt
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    private fun checkTaskCompletion(
+        tasks: List<Task>,
+        exercises: List<ExerciseSessionInfo>
+    ): List<SportTask> {
+        return tasks.mapNotNull { task ->
+            val activity = (task as SportTask).activity
+            var duration = task.duration ?: Duration.ZERO
```
klaidoshka, line 58: In case both duration and distance are null, but we have some exercise data, this could lead the task without set duration/distance to be marked as completed, right?

I think this should be somewhat adjusted for that, if that's the case. Because for now duration/distance are mandatory when creating a task, but I really believe they should not be, since user may want to assign to themself a general task, which may take unknown time `Task#duration` and which should begin at unknown time `Task#time` - just something to do along the day.

app/src/main/java/com/pvp/app/service/TaskAutocompleteService.kt
```diff
@@ -0,0 +1,192 @@
...
+            null
+        }
+    }
+
+    private fun createNotification(): Notification {
+        return NotificationCompat.Builder(
+            this,
+            CHANNEL_ID
+        )
+            .setContentTitle("Task Processing")
```
klaidoshka, line 106: If using NotificationChannel instead of const string, you could use NotificationChannel#id here instead.

app/src/main/java/com/pvp/app/worker/TaskAutocompleteWorker.kt
```diff
@@ -0,0 +1,41 @@
...
+        const val WORKER_NAME = "TaskAutocompleteWorker"
+    }
+
+    override suspend fun doWork(): Result {
+        val context = applicationContext
+
+        Intent(
+            context,
+            TaskAutocompleteService::class.java
+        ).also { intent ->
```
klaidoshka, line 32: Split

app/src/main/java/com/pvp/app/worker/TaskAutocompleteWorker.kt
```diff
@@ -0,0 +1,41 @@
...
+    context,
+    workerParams
+) {
+    companion object {
+
+        const val WORKER_NAME = "TaskAutocompleteWorker"
+    }
+
+    override suspend fun doWork(): Result {
+        val context = applicationContext
```
klaidoshka, line 27: Is this _really_ necessary?

KhajiitLikeToSneak (approved):

klaidoshka (approved):

PR #75, opened 2024-04-11
PVP-231 Upon clicking a faded side day-card, navigate to that day

youreondrugs:
surrounding day cards are now clickable. clicking on them navigates to that day card.


klaidoshka (commented):
Seems too complex than it should be. I believe you can access the statePager within onClick lambda in DayCard component, so, additional variable with LaunchedEffect instantly would become redundant.

Also, in case the clickEnabled property in DayCard is always true (in all places this component is used), it should be removed and component should not handle it then.

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarWeeklyScreen.kt
```diff
@@ -63,14 +67,20 @@ fun Week(
...
-                    if (tasksFiltered.isEmpty()) {
-                        stateDialog = true
-                    } else {
-                        stateShowCards = !stateShowCards
+                    if (date != dates[page]) {
+                        scope.launch {
+                            statePager.scrollToPage(page)
+                        }
+                    }
+                    else {
```
klaidoshka, line 78: This is not our used style for conditional statements. `else` should be after closing `if` body bracket, in the same line. Ctrl+Alt+L

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarWeeklyScreen.kt
```diff
@@ -8,9 +8,11 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.pager.HorizontalPager
 import androidx.compose.foundation.pager.rememberPagerState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
```
klaidoshka, line 11: Is it used? Ctrl+Alt+O

klaidoshka (approved):

vytjar (approved):

PR #77, opened 2024-04-14
PVP-212

KhajiitLikeToSneak:
Implemented friends system.
- Users are able to add friends via their email in the “Friends“ screen. 
- Friend list is visible in this screen.
- User is able to cancel sent request.
- Another user is able to accept/deny received request.
- Applied design created in PVP-218 and PVP-221.


app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        if (friendEmail in friendObject.friends) {
+            return "$friendEmail is already your friend"
+        }
+
+        if (friendEmail in friendObject.sentRequests) {
+            return "Friend request already sent to $friendEmail"
+        }
+        if (friendEmail in friendObject.receivedRequests) {
+            return "Friend request already received from $friendEmail"
```
klaidoshka, line 97: We are trying to send a request here, right? User may not understand what this means. We should navigate the user to the right actions or do it automatically (better) by accepting the request that is already received from before.

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        val friendNew = friend.copy(receivedRequests = friend.receivedRequests + userEmail)
+
+        merge(friendNew, friendEmail)
```
klaidoshka, line 102: Split

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        val friendObjectNew =
+            friendObject.copy(sentRequests = friendObject.sentRequests + friendEmail)
```
klaidoshka, line 105: Try to fit into the same line, split after parameters start `.copy(<...>`.

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        merge(friendObjectNew, userEmail)
```
klaidoshka, line 107: Split.

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        return "Friend request sent!"
+    }
+
+    override suspend fun acceptFriendRequest(
+        friendObject: FriendObject,
+        userEmail: String,
+        friendEmail: String
+    ): String {
+        val friend = get(friendEmail).first() ?: return "Friend not found"
```
klaidoshka, line 117: Split `first` (and `?: return`, if that looks suitable to you).

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        val friendNew = friend.copy(
+            sentRequests = friend.sentRequests - userEmail,
+            friends = friend.friends + userEmail
+        )
+
+        merge(friendNew, friendEmail)
```
klaidoshka, line 124: Split

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        val userNew = friendObject.copy(
+            receivedRequests = friendObject.receivedRequests - friendEmail,
+            friends = friendObject.friends + friendEmail
+        )
+
+        merge(userNew, userEmail)
```
klaidoshka, line 131: Split

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        return "Friend request accepted!"
+    }
+
+    override suspend fun denyFriendRequest(
+        friendObject: FriendObject,
+        userEmail: String,
+        friendEmail: String
+    ): String {
+        val friend = get(friendEmail).first() ?: return "Friend not found"
```
klaidoshka, line 141: Split

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        val friendNew = friend.copy(sentRequests = friend.sentRequests - userEmail)
+
+        merge(friendNew, friendEmail)
```
klaidoshka, line 145: Split

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        val userNew =
```
klaidoshka, line 147: Fit by splitting

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+            friendObject.copy(receivedRequests = friendObject.receivedRequests - friendEmail)
+
+        merge(userNew, userEmail)
```
klaidoshka, line 150: Split

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        return "Friend request denied!"
+    }
+
+    override suspend fun cancelSentRequest(
+        friendObject: FriendObject,
+        userEmail: String,
+        friendEmail: String
+    ): String {
+        val friend = get(friendEmail).first() ?: return "Friend not found"
```
klaidoshka, line 160: Split

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        val friendNew = friend.copy(receivedRequests = friend.receivedRequests - userEmail)
+
+        merge(friendNew, friendEmail)
```
klaidoshka, line 164: Split

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        val userNew = friendObject.copy(sentRequests = friendObject.sentRequests - friendEmail)
+
+        merge(userNew, userEmail)
```
klaidoshka, line 168: Split

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsScreen.kt
```diff
@@ -0,0 +1,435 @@
...
+    LaunchedEffect(toastMessage) {
+        toastMessage?.let {
+            Toast
+                .makeText(
+                    context,
+                    it,
+                    Toast.LENGTH_SHORT
+                )
+                .show()
+            model.toastMessage.value = null
```
klaidoshka, line 84: Separate

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsScreen.kt
```diff
@@ -0,0 +1,435 @@
...
+        }
+    }
+
+    LaunchedEffect(key1 = Unit) {
```
klaidoshka, line 88: Could drop the `key1 =` part, it is the first parameter, will be cleaner to read.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsScreen.kt
```diff
@@ -0,0 +1,435 @@
...
+            onClick = { showRequests.value = true },
+            modifier = Modifier.size(
+                width = 160.dp,
+                height = 35.dp
+            ),
+            content = ButtonContent.Text("Requests")
+        )
+
+        RowButton(
+            onClick = { /* TODO */ },
```
klaidoshka, line 265: What is this for? Comment doesn't really help to understand what has to be implemented.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsScreen.kt
```diff
@@ -0,0 +1,435 @@
...
+                modifier = Modifier.fillMaxWidth()
+            )
+        } else {
+            for (request in requests) {
+                Row(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .height(30.dp)
+                        .padding(
+                            vertical = 2.dp
```
klaidoshka, line 392: Could close the brackets for single parameter

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsScreen.kt
```diff
@@ -0,0 +1,435 @@
...
+private fun RequestList(
+    requests: List<String>,
+    requestTitle: String,
+    acceptAction: (String) -> Unit,
+    denyAction: (String) -> Unit
+) {
+    val scrollState = rememberScrollState()
+
+    Column(
+        modifier = Modifier.verticalScroll(scrollState)
```
klaidoshka, line 376: Could close the brackets for single parameter

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
```
klaidoshka, line 20: Redundant line

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+
+@HiltViewModel
+class FriendsViewModel @Inject constructor(
+    private val friendService: FriendService,
+    private val userService: UserService
+) : ViewModel() {
+    val toastMessage = mutableStateOf<String?>(null)
```
klaidoshka, line 27: Requires single empty line above. Also try to separate from other two paired values with another empty line below. Think of them as one single multi-line unit.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+        } ?: flowOf(null)
+    }.stateIn(
+        viewModelScope,
+        started = SharingStarted.WhileSubscribed(5000),
+        initialValue = null
+    )
+
+    init {
+        viewModelScope.launch {
+            user.collect { user ->
```
klaidoshka, line 50: Do we really want to keep trying to create friend object for the user, whenever user object's data changes only to fail? I think this should be done only once, without repeats.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+            friendService.createFriendObject(email)
+        }
+    }
+
+    fun addFriend(friendEmail: String){
+        viewModelScope.launch {
+            if (friendEmail.isEmpty()) {
+                toastMessage.value = "Please enter an email"
+                _isRequestSent.value = false
+                return@launch
```
klaidoshka, line 69: Separate return

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+            }
+
+            val userEmail = user.value?.email ?: return@launch
+            val friendObject = friendService.get(userEmail).firstOrNull() ?: return@launch
```
klaidoshka, line 73: Variables should be split on chained method calls. After split, values will become multi-line and will have to be separated from above/below.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+            val friendUser = userService.get(friendEmail).firstOrNull()
+
+            if (friendUser == null) {
+                toastMessage.value = "User with email $friendEmail does not exist"
+                _isRequestSent.value = false
+                return@launch
```
klaidoshka, line 79: Separate return

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+            }
+
+            friendService.createFriendObject(friendEmail)
+
+            val toastMessageValue = friendService.addFriend(
+                friendObject,
+                userEmail,
+                friendEmail
+            )
+            toastMessage.value = toastMessageValue
```
klaidoshka, line 89: Separate from variable initialization.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+            _isRequestSent.value = toastMessageValue == "Friend request sent!"
+        }
+    }
+
+    fun acceptFriendRequest(friendEmail: String) {
+        viewModelScope.launch {
+            val userEmail = user.value?.email ?: return@launch
+            val friendObject = friendService.get(userEmail).firstOrNull() ?: return@launch
```
klaidoshka, line 97: Split, separate after (above)

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+                userEmail,
+                friendEmail
+            )
+        }
+    }
+
+    fun denyFriendRequest(friendEmail: String) {
+        viewModelScope.launch {
+            val userEmail = user.value?.email ?: return@launch
+            val friendObject = friendService.get(userEmail).firstOrNull() ?: return@launch
```
klaidoshka, line 110: Same as above

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+                userEmail,
+                friendEmail
+            )
+        }
+    }
+
+    fun cancelSentRequest(friendEmail: String) {
+        viewModelScope.launch {
+            val userEmail = user.value?.email ?: return@launch
+            val friendObject = friendService.get(userEmail).firstOrNull() ?: return@launch
```
klaidoshka, line 123: Same as above

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    fun getFriendAvatar(friendEmail: String): ImageBitmap {
+        var avatar: ImageBitmap? = null
+
+        viewModelScope.launch {
+            avatar = userService.resolveAvatar(friendEmail)
+        }
+
+        return avatar ?: ImageBitmap(0, 0)
```
klaidoshka, line 140: ImageBitmap of 0px width/height is going to throw an error within certain chained constructor, I believe.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,161 @@
...
+    val user = userService.user.stateIn(
+        viewModelScope,
+        started = SharingStarted.WhileSubscribed(5000),
+        initialValue = null
+    )
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val friendObject = user.flatMapLatest { user ->
+        user?.email?.let { email ->
+            friendService.create(email)
```
klaidoshka, line 41: Every time user changes, we are still trying to create a friend object. To me it looks that getting nullable user in this ViewModel is hardly possible, since the route is locked until `User` object becomes not-null. So you could just call `user.first()!!` or `user.first()`, but make `Flow<User?>` to filterNotNull(mapNotNull?) and then create its friend object within ViewModel's init block (or other private function, but the idea is that it would be called in the init block).

However. I think it would be best to create the friend object whenever a new user is created after register. What you are doing now is probably to support old users so that they get their objects created. Well, you could write a small piece of code that would go through every user and create their friend objects upon application start, then remove that piece of code and leave it all for the register process by modifying it a bit.

You choose how you want to handle this, if you want, even leave this thing as it is, it's not like it takes 1 or more seconds to complete. But still, small resource waste, especially if we had to pay for our database reads. 😄

Also, two different purpose methods should be separated.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,161 @@
...
+    val isRequestSent = _isRequestSent.asStateFlow()
+
+    val user = userService.user.stateIn(
+        viewModelScope,
+        started = SharingStarted.WhileSubscribed(5000),
+        initialValue = null
+    )
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val friendObject = user.flatMapLatest { user ->
```
klaidoshka, line 39: flatMapLatest, stateIn should be split.

vytjar (approved):

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -0,0 +1,161 @@
...
+            val toastMessageValue = friendService.addFriend(
+                friendObject,
+                email,
+                friendEmail
+            )
+
+            toastMessage.value = toastMessageValue
+            _isRequestSent.value = (
+                    toastMessageValue == "Friend request sent!" ||
+                    toastMessageValue == "Friend request accepted!"
```
klaidoshka, line 96: If ViewModel's code is hard dependent on what string is returned from the unknown Service implementation, then it should be somehow controlled a little more, so that the changes in the Service implementations would also take place in ViewModel's implementation.

This looks very bad practice, so if you don't want to change this with code, then at least put comments onto the Service interface appropriate methods indicating that returned string is strictly associated with how ViewModel works. So if any changes to the default implementation are made, changes in the ViewModel should also be made.

klaidoshka (approved):

PR #78, opened 2024-04-16
PVP-222 Add hydration notifications toggler

AurisTFG:


app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -190,13 +221,22 @@ fun CategoryRow(
...
-    value: String
+    title: String,
+    value: T,
+    isEnabled: Boolean = true
 ) {
+    var textColor = MaterialTheme.colorScheme.onPrimary
+    var backgroundColor = MaterialTheme.colorScheme.primary
+
+    if (!isEnabled) {
+        textColor = textColor.copy(alpha = 0.5f)
```
klaidoshka, line 236:
```kotlin
.lighten()
.darken()
```
Our extensions exist that do exactly this and accept ratio in case you want to provide something else than 0.5

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -207,61 +247,98 @@ fun SettingCard(
...
-        }
+                Spacer(modifier = Modifier.size(8.dp))
 
-        if (state) {
-            Dialog(
-                onDismissRequest = {
-                    onEdit()
+                Icon(
+                    tint = textColor,
+                    contentDescription = "Open dialog to edit a setting",
```
klaidoshka, line 327: This should refer to which setting more as all the other accessibility support properties. In case user is blind and is using sound device's help to navigate, each setting icon will just say "Open a dialog to edit a setting". But what setting? And is this icon or something else?

If we are providing accessibility support, let's make it half decent or not provide it at all.

PR #79, opened 2024-04-16
PVP-202 Add Reset to Default button in Settings Screen

AurisTFG:


app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -193,6 +195,31 @@ fun SettingsScreen(
         SettingHydrationNotificationToggle(model)
 
         SettingCupVolumeMl(model)
+
+        Column(
+            modifier = Modifier.fillMaxSize(),
+            horizontalAlignment = Alignment.CenterHorizontally
+        ) {
+            ButtonConfirm(
```
klaidoshka, line 203: I think that it would be nice if this button would have similar style as profile's `Delete Account` button.

klaidoshka (approved):

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
klaidoshka: Actually, please also transfer the whole reset button container out of the SettingsScreen component so that it would become more readable (smaller) with only component names, which tell the functionality.

vytjar (approved):

PR #80, opened 2024-04-16
PVP-201: Allow changing theme in the settings

youreondrugs:


app/src/main/java/com/pvp/app/Activity.kt
```diff
@@ -31,7 +33,8 @@ class Activity : ComponentActivity() {
         prepareRoutes()
 
         setContent {
-            CalendarTheme {
+            val themeViewModel: ThemeViewModel = hiltViewModel()
+            CalendarTheme (model = themeViewModel) {
```
AurisTFG, line 37: Empty line needed

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -173,6 +175,118 @@ private fun SettingHydrationNotificationToggle(
...
+private fun SettingApplicationTheme(
+    model: SettingsViewModel = hiltViewModel()
+) {
+    val themeValue by model
+        .get(Setting.ApplicationTheme)
+        .collectAsStateWithLifecycle()
+
+    val state = rememberPickerState(initialValue = themeValue)
+
+    val valueText = when (themeValue) {
```
AurisTFG, line 188: maybe use Enum here?

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -173,6 +175,118 @@ private fun SettingHydrationNotificationToggle(
...
+                    .padding(32.dp)
+            ) {
+                Text(
+                    style = MaterialTheme.typography.titleLarge,
+                    text = "Select Application Theme"
+                )
+
+                Spacer(modifier = Modifier.size(16.dp))
+
+                Row(
```
AurisTFG, line 215: this whole Row could be wrapped in a function to avoid repeating same code three times

app/src/main/java/com/pvp/app/ui/theme/Theme.kt
```diff
@@ -27,20 +34,32 @@ private val LightColorScheme = lightColorScheme(
     tertiary = LightTertiary
 )
 
+@HiltViewModel
+class ThemeViewModel @Inject constructor(
```
AurisTFG, line 38: Maybe move this to a separate file ThemeViewModel.kt?

app/src/main/java/com/pvp/app/ui/theme/Theme.kt
```diff
@@ -27,20 +34,32 @@ private val LightColorScheme = lightColorScheme(
...
+    private val settingService: SettingService
+) : ViewModel() {
+    fun getSetting(): Flow<Int> {
+        return settingService.get(Setting.ApplicationTheme)
+    }
+}
+
 @Composable
 fun CalendarTheme(
+    model: ThemeViewModel,
```
AurisTFG, line 48: model: ThemeViewModel = hiltViewModel() to automatically inject the viewmodel (therefore could also remove ThemeViewModel mentions from Activity.kt completely)

app/src/main/java/com/pvp/app/ui/theme/Theme.kt
```diff
@@ -27,20 +34,32 @@ private val LightColorScheme = lightColorScheme(
...
     darkTheme: Boolean = isSystemInDarkTheme(),
```
AurisTFG, line 49: since this argument is not really used from the outside ever, maybe it could be removed and then the function call could just be moved into the if condition

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -173,6 +175,118 @@ private fun SettingHydrationNotificationToggle(
...
+                    Modifier
+                        .fillMaxWidth()
+                        .padding(horizontal = 16.dp)
+                ) {
+                    RadioButton(
```
AurisTFG, line 220: the RadioButton and Text couple dont seem to be vertically aligned on my end.

AurisTFG (approved):

app/src/main/java/com/pvp/app/Activity.kt
klaidoshka: Also redundant space

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
klaidoshka: A little hard to understand, but if this is the only place we will handle the values, it could stay as it is. In case we had to handle the value in multiple (different) places, then we really should change it to enum.

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -173,6 +175,118 @@ private fun SettingHydrationNotificationToggle(
...
+    val themeValue by model
+        .get(Setting.ApplicationTheme)
+        .collectAsStateWithLifecycle()
+
+    val state = rememberPickerState(initialValue = themeValue)
+
+    val valueText = when (themeValue) {
+        1 -> "Dark"
+        2 -> "Light"
+        3 -> "Dynamic"
```
klaidoshka, line 191: Dynamic can also be dark or light. You should allow choosing those too. But it becomes repetitive then. So different solution could be made: a `switch` for dynamic and a `checkbox` for dark, light themes. We should also have "auto", which checks device's theme and uses it, in case we can't find device's theme (old device), we would fallback into light theme.

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -193,6 +307,13 @@ fun SettingsScreen(
         SettingHydrationNotificationToggle(model)
 
         SettingCupVolumeMl(model)
+
+        CategoryRow(
+            icon = Icons.Outlined.Snowboarding,
+            title = "Theme"
```
klaidoshka, line 313: Maybe `Appearance` instead of `Theme`.
And [Style](https://mui.com/material-ui/material-icons/?theme=Outlined&query=style&selected=StyleOutlined) icon, if it exists in Android.

app/src/main/java/com/pvp/app/ui/theme/Theme.kt
```diff
@@ -27,20 +34,32 @@ private val LightColorScheme = lightColorScheme(
     tertiary = LightTertiary
 )
 
+@HiltViewModel
+class ThemeViewModel @Inject constructor(
+    private val settingService: SettingService
+) : ViewModel() {
+    fun getSetting(): Flow<Int> {
```
klaidoshka, line 41: Empty line should be added.

klaidoshka (changes requested):

klaidoshka:
You forgot to clean-up things. `Ctrl+Alt+L` and `Ctrl+Alt+O` would help out.

- We are always showing wrong **current** theme value, which defaults to 1.
image
- We should show text, not numbers. Users have no clue what 0, 1, 2 means.
- `Auto` is not working, it just always provides Light theme, no matter if we switch to dark theme in our Android OS.

Just my opinion:
We could drop `Set` from the `Set Application Theme` title. Not necessary, but I think it would be more fit.
Other settings could also get their `Set` discarded from the title, **BUT** that would be out of the scope.

app/src/main/java/com/pvp/app/ui/theme/Theme.kt
```diff
@@ -29,19 +36,31 @@ private val LightColorScheme = lightColorScheme(
...
         dynamicColor && Build.VERSION.SDK_INT >= Build.VERSION_CODES.S -> {
             val context = LocalContext.current
-
-            if (darkTheme) dynamicDarkColorScheme(context) else dynamicLightColorScheme(context)
+            if (themeValue == 0) {
+                dynamicDarkColorScheme(context)
+            } else if (themeValue == 1) {
+                dynamicLightColorScheme(context)
+            } else {
+                if (darkTheme) dynamicLightColorScheme(context)
```
klaidoshka, line 53: Wrong return

app/src/main/java/com/pvp/app/ui/theme/Theme.kt
```diff
@@ -29,19 +36,31 @@ private val LightColorScheme = lightColorScheme(
...
+                else dynamicLightColorScheme(context)
+            }
         }
 
-        darkTheme -> DarkColorScheme
-        else -> LightColorScheme
+        themeValue == 0 -> DarkColorScheme
+        themeValue == 1 -> LightColorScheme
+        else -> {
+            if (darkTheme) LightColorScheme
```
klaidoshka, line 61: Same as above

klaidoshka (approved):
Apart possible clean up, looks good.

PR #81, opened 2024-04-16
PVP-204 Add Delete Account button in Profile Screen

AurisTFG:


app/src/main/java/com/pvp/app/ui/screen/profile/ProfileViewModel.kt
```diff
@@ -42,6 +51,29 @@ class ProfileViewModel @Inject constructor(
             )
         )
 
+    suspend fun deleteAccount(): Result<Boolean> {
+        return withContext(Dispatchers.IO) {
+            try {
+                val email = state.value.user.email
+
+                taskService.get(email).first().forEach { task ->
```
klaidoshka, line 59: Split

app/src/main/java/com/pvp/app/ui/screen/profile/ProfileViewModel.kt
```diff
@@ -42,6 +51,29 @@ class ProfileViewModel @Inject constructor(
...
+                    taskService.remove(task)
+                }
+
+                friendService.remove(email)
+
+                userService.remove(email)
+
+                authenticationService.user.first()?.delete()?.await()
```
klaidoshka, line 67: Split

app/src/main/java/com/pvp/app/ui/screen/profile/ProfileViewModel.kt
```diff
@@ -42,6 +51,29 @@ class ProfileViewModel @Inject constructor(
             )
         )
 
+    suspend fun deleteAccount(): Result<Boolean> {
```
klaidoshka, line 54: We are pretty much handling the whole logic here, which I tend to think shouldn't be done in the ViewModel. ViewModel should act as Controller would in MVC architecture - controlling what is the logic of process flow pretty much, what is called and on what conditions.

I believe that user removing logic should be handled in the AuthenticationService.

klaidoshka (approved):

vytjar (approved):

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -56,16 +56,39 @@ class FriendServiceImpl @Inject constructor(
...
-                    .collection(identifier)
-                    .document(email)
+            .collection(identifier)
+            .document(email)
+            .delete()
+            .await()
 
-                transaction.delete(document)
-            }
+        database
```
klaidoshka, line 64: When removing friend object, we are removing the initial object and then iterating through **every** other friend object, no matter if they are related or not, then updating each. At least it seems so.

This is very redundant iteration, when we could take the initial object and check what friends it stores, then map to those specific friends in iteration.

PR #83, opened 2024-04-17
PVP-208 PVP-211 PVP-215 PVP-216 Add decorations feature

klaidoshka:
# PVP-208 PVP-211 PVP-215 PVP-216
# Changes
- Added Coil library for images handling
- Removed default avatar .svg from project files, now using urls of different avatar parts from the database
- Added ImageService that handles image requests with Coil
- Added DecorationService for decorations handling - fetching, applying, checking, purchasing, etc.
- Decoration model consists of decoration type, title, price and two image urls: layer, representative. Layer is used for applying decorations onto images, while representative is for displaying decoration within shop.
- `Components.kt` update:
  - Moved `Dialog` out of the `Buttons.kt` into `Components.kt`, made it for public use.
  - Added AsyncImage wrapper for Coil, for easier use
  - Added FoldableContent wrapper that shows content when it is not folded. Folding happens on clicking the header.
- `Modifiers.kt` was modified a bit to support more parameters in `Modifier#underline`, which was used at first, but then dropped.
- Fixed padding problems in authenticated layout, when not in Profile screen (by swapping to the right). 

Decorations page supports purchasing and selecting decorations. Decorations are applied in real-time, user can see the changes on showcased avatar and also in the header, other places where avatar is showcased. Some errors are handled and shown to the user via a toast.

### I highly suggest compiling this branch and test it. To find any missed or unnecessary spots.

# PVP-223
# Changes
Implemented "streak" functionality, that rewards user for using the app everyday.

- Users receive points and experience for log in everyday. The amount of points and experience the user receive increases for each day user logs in in a row. Maximum is reached at day 7.

Current values:
day 1: 1 point, 10 xp
day 2: 3 points, 20 xp
day 7: 13 points, 70 xp
day 8: 13 points, 70 xp

- Additional rewards are set in the database in the rewards collection. "Streak" value is used as the id. User can receive additional points, experience or a decoration (if not received yet)
- When user received a reward, a dialog opens up informing them about what reward they received.

Currently it looks like this:

image

The design should probably be reworked in the future.


AurisTFG (approved):

app/src/main/java/com/pvp/app/ui/screen/decoration/Models.kt
```diff
@@ -0,0 +1,89 @@
...
+    val avatar: ImageBitmap = ImageBitmap(
+        1,
+        1
+    ),
+    val holders: List<DecorationHolder> = emptyList(),
+    val workState: WorkState = WorkState.NoOperation,
+    val user: User = User()
+)
+
+sealed class WorkState {
```
klaidoshka, line 27: Rename to DecorationScreenState? Adjust method name in VM.

app/src/main/java/com/pvp/app/ui/screen/layout/LayoutScreenAuthenticated.kt
```diff
@@ -286,9 +316,9 @@ fun LayoutScreenAuthenticated(
         },
         drawerState = stateDrawer
     ) {
-        var isOpen by remember { mutableStateOf(false) }
+        var isTaskDialogOpen by remember { mutableStateOf(false) }
 
-        val toggleDialog = remember { { isOpen = !isOpen } }
+        val toggleTaskDialog = remember { { isTaskDialogOpen = !isTaskDialogOpen } }
```
klaidoshka, line 321: Same as above

app/src/main/java/com/pvp/app/ui/screen/layout/LayoutScreenAuthenticated.kt
```diff
@@ -345,11 +375,106 @@ fun LayoutScreenAuthenticated(
...
+    isOpen: Boolean,
+    onClose: () -> Unit
+) {
+    if (!isOpen) {
+        return
+    }
+
+    Dialog(
+        onDismissRequest = onClose
+    ) {
```
klaidoshka, line 399: Put in one line

app/src/main/java/com/pvp/app/ui/screen/layout/LayoutScreenAuthenticated.kt
```diff
@@ -345,11 +375,106 @@ fun LayoutScreenAuthenticated(
...
+                DecorationCard(decoration = it)
+            }
+        }
+    }
+}
+
+@Composable
+private fun DecorationCard(
+    decoration: Decoration
+) {
```
klaidoshka, line 454: Put in one line

KhajiitLikeToSneak (approved):

PR #89, opened 2024-04-19
PVP-214

KhajiitLikeToSneak:
- Added friend information display.
- Refactored friends model to contain 'friends since when' value.
- Added an ability to remove friend in friend information dialog.
- Added another getAvatar function in DecorationService from presentation branch, to display friends avatars.
- Separated a lot of code to smaller functions to increase readability.


app/src/main/java/com/pvp/app/ui/screen/friends/FriendsScreen.kt
```diff
@@ -441,4 +738,114 @@ private fun RequestList(
...
+    )
+}
+
+@Composable
+private fun CustomDialog(
+    content: @Composable () -> Unit,
+    title: @Composable () -> Unit,
+    buttonContentConfirm: @Composable RowScope.() -> Unit,
+    buttonContentDismiss: @Composable RowScope.() -> Unit,
+    onConfirm: () -> Unit,
```
klaidoshka, line 806: Unused

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -180,6 +194,55 @@ class FriendsViewModel @Inject constructor(
...
+                .firstOrNull()
+
+            val mutualFriendsList =
+                friendObject?.friends?.map { it.email }
+                    ?.intersect(userFriends.value.map { it.user.email }
+                        .toSet())
+
+            if (mutualFriendsList != null) {
+                val mutualFriendsUsers = mutualFriendsList.mapNotNull {
+                    val user = userService.get(it).firstOrNull()
```
klaidoshka, line 211: Split

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -180,6 +194,55 @@ class FriendsViewModel @Inject constructor(
...
+
+    fun getMutualFriends(friendEmail: String) {
+        viewModelScope.launch {
+            val friendObject = friendService
+                .get(friendEmail)
+                .firstOrNull()
+
+            val mutualFriendsList =
+                friendObject?.friends?.map { it.email }
+                    ?.intersect(userFriends.value.map { it.user.email }
```
klaidoshka, line 206: Split

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -180,6 +194,55 @@ class FriendsViewModel @Inject constructor(
...
+                        .toSet())
+
+            if (mutualFriendsList != null) {
+                val mutualFriendsUsers = mutualFriendsList.mapNotNull {
+                    val user = userService.get(it).firstOrNull()
+
+                    user?.let {
+                        FriendEntry(
+                            avatar = decorationService.getAvatar(user),
```
klaidoshka, line 215: This is not using flow (your added decoration method). So if any changes are being made to the decorations, they will not trigger update on this avatar.

It is not ideal, but it is also only shown when user info is clicked, right? So, since it is not a place that is seen 24/7, this should be fine as it is, since on each info open it should recompute.

_Also, decorations being updated is hardly going to happen, but it can._

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -180,6 +194,55 @@ class FriendsViewModel @Inject constructor(
...
+                mutualFriends.value = mutualFriendsUsers
+            } else {
+                mutualFriends.value = emptyList()
+            }
+        }
+    }
+
+    fun tasksCompleted(friendEmail: String) {
+        viewModelScope.launch {
+            val tasks = taskService.get(friendEmail).firstOrNull() ?: emptyList()
```
klaidoshka, line 229: Separate, split

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -180,6 +194,55 @@ class FriendsViewModel @Inject constructor(
...
+            tasksCompleted.value = tasks.count {
+                it.isCompleted && it.date in LocalDate.now().minusDays(7)..LocalDate.now()
```
klaidoshka, line 231: Should be split. I wonder how it will look.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -180,6 +194,55 @@ class FriendsViewModel @Inject constructor(
...
+    }
+
+    fun removeFriend(friendEmail: String) {
+        viewModelScope.launch {
+            friendService.removeFriend(
+                userFriendObject.value,
+                user.value.email,
+                friendEmail
+            )
+            toastMessage.value = "Friend removed successfully!"
```
klaidoshka, line 243: Separate

klaidoshka (approved):

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -180,6 +194,55 @@ class FriendsViewModel @Inject constructor(
...
     }
+
+    fun getMutualFriends(friendEmail: String) {
+        viewModelScope.launch {
+            val friendObject = friendService
+                .get(friendEmail)
+                .firstOrNull()
+
+            val mutualFriendsList =
+                friendObject?.friends?.map { it.email }
```
klaidoshka, line 205: **All cases:**
- If there are properties in the chained call, they go within the same line (until no longer fits).
- If there is one method in whole chained call, it stays within the same line (until no longer fits).
- If there are multiple methods in whole chained call, each method is split into new line. If there are property calls in between method calls, these properties stay in the same line as the last called method.

**Here would be:**
```kotlin
val friends = friendObject?.friends
    ?.map { it.email }
    ?.intersect(
        userFriends.value
            .map { it.user.email }
            .toSet()
    )
```

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsViewModel.kt
```diff
@@ -180,6 +194,69 @@ class FriendsViewModel @Inject constructor(
...
+                        .firstOrNull()
+
+                    user?.let {
+                        FriendEntry(
+                            avatar = decorationService.getAvatar(user),
+                            user = user
+                        )
+                    }
+                }
+                mutualFriends.value = mutualFriendsUsers
```
klaidoshka, line 224: Separate from multi-line variable block

app/src/main/java/com/pvp/app/service/FriendServiceImpl.kt
```diff
@@ -100,6 +101,32 @@ class FriendServiceImpl @Inject constructor(
...
+    ) {
+        val friendObjectFriendsNew = friendObject.friends.filter { it.email != friendEmail }
+        val friendObjectNew = friendObject.copy(friends = friendObjectFriendsNew)
+
+        merge(
+            friendObjectNew,
+            friendObjectEmail
+        )
+
+        val friendNew = get(friendEmail).first()
```
klaidoshka, line 117: Split, then separation will be required.

vytjar (approved):

PR #95, opened 2024-04-25
PVP-243: Add autocomplete indicator near tasks that are likely to be autocompleted

youreondrugs:
after selecting an activity in the task form a note appears above the create button indicating that the task will likely be autocompleted if the activity supports distance metrics:

image

the note disappears after changing to a non-distance activity or to a different type of task (general or meal).

### DIFF
--- a/app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsCreate.kt
+++ b/app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsCreate.kt
@@ -143,6 +143,7 @@ fun TaskCreateForm(
     var duration by remember { mutableStateOf(Duration.ofMinutes(0)) }
     var reminderTime by remember { mutableStateOf<Duration?>(null) }
     var activity by remember { mutableStateOf(SportActivity.Walking) }
+    var supportsDistanceMetrics by remember { mutableStateOf(false) }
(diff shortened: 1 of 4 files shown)


klaidoshka (changes requested):

klaidoshka:
It would be nice if this tooltip was also added to the task card component (in day's task list preview).

klaidoshka:
How is it going?

app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsCreate.kt
```diff
@@ -143,6 +143,7 @@ fun TaskCreateForm(
     var duration by remember { mutableStateOf(Duration.ofMinutes(0)) }
     var reminderTime by remember { mutableStateOf<Duration?>(null) }
     var activity by remember { mutableStateOf(SportActivity.Walking) }
+    //var supportsDistanceMetrics by remember { mutableStateOf(false) }
```
klaidoshka, line 146: These are your changes and you aren't using this variable at all. Instead of commenting out and leaving them as is, you should completely remove them (4 lines in total) to not clutter code base with redundant comments (and/or code).

klaidoshka (approved):

PR #96, opened 2024-04-28
PVP-235

vytjar:
Implemented goal creation and view

- User can create weekly and monthly goals for activities that we deem likely to be autocompleted
- When user is creating a walking goal, they can choose between steps and distance. Otherwise they can only choose distance
- When creating a goal, the start date is always automatically chosen as today, if user choose to create a weekly goal - end date is 7 days from today, monthly - 30 days.
- Viewer can change dates for weekly/monthly goals they're viewing. Goal is shown if his end date or start date falls between the shown range.
- Once user creates a goal, it can not be deleted

Here's how the implementation currently looks:

### DIFF
--- a/app/src/main/java/com/pvp/app/api/GoalService.kt
+++ b/app/src/main/java/com/pvp/app/api/GoalService.kt
@@ -1,0 +1,47 @@
+package com.pvp.app.api
+
+import com.pvp.app.model.Goal
+import com.pvp.app.model.SportActivity
+import kotlinx.coroutines.flow.Flow
+import java.time.LocalDate
+
+interface GoalService: DocumentsCollection {
+
+    override val identifier: String
+        get() = "goals"
+
+    /**
+     * Creates a new goal in the database.
+     */
+    suspend fun create(
+        activity: SportActivity,
+        endDate: LocalDate,
+        email: String,
+        goal: Double,
+        monthly: Boolean,
+        startDate: LocalDate,
+        steps: Boolean
+    )
+
+    suspend fun claim(
+        goal: Goal
+    )
+
+    /**
+     * Gets all goals for the specified email.
+     */
+    suspend fun get(email: String): Flow<List<Goal>>
+
+    /**
+     * Gets all goals for the specified email and start date.
+     */
+    suspend fun get(
+        email: String,
+        startDate: LocalDate
+    ): Flow<List<Goal>>
+
+    /**
+     * Updates the goal in the database.
+     */
+    suspend fun update(goal: Goal)
+}
--- a/app/src/main/java/com/pvp/app/api/PointService.kt
+++ b/app/src/main/java/com/pvp/app/api/PointService.kt
@@ -1,5 +1,6 @@
 package com.pvp.app.api
 
+import com.pvp.app.model.Goal
 import com.pvp.app.model.Task
 import java.time.LocalDate
 
@@ -31,6 +32,16 @@ interface PointService {
     ): Int {
         return tasks.sumOf { calculate(it) }
     }
+
+    /**
+     * Calculate the points of a goal. Points are determined by the goal type and the attributes of
+     * the goal. User status can also be taken into account, hence, [Goal.email] is required.
+     *
+     * @param goal goal to calculate points for
+     *
+     * @return points of the goal
+     */
+    suspend fun calculate(goal: Goal): Int
 
     /**
      * Deduct points from current user for not completing tasks.
--- a/app/src/main/java/com/pvp/app/service/GoalServiceImpl.kt
+++ b/app/src/main/java/com/pvp/app/service/GoalServiceImpl.kt
@@ -1,0 +1,165 @@
+package com.pvp.app.service
+
+import com.google.firebase.firestore.FirebaseFirestore
+import com.google.firebase.firestore.snapshots
+import com.pvp.app.api.ExperienceService
+import com.pvp.app.api.GoalService
+import com.pvp.app.api.PointService
+import com.pvp.app.api.UserService
+import com.pvp.app.common.JsonUtil.JSON
+import com.pvp.app.common.JsonUtil.toJsonElement
+import com.pvp.app.common.JsonUtil.toPrimitivesMap
+import com.pvp.app.model.Goal
+import com.pvp.app.model.Points
+import com.pvp.app.model.SportActivity
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.tasks.await
+import kotlinx.serialization.json.decodeFromJsonElement
+import kotlinx.serialization.json.encodeToJsonElement
+import java.time.LocalDate
+import java.time.LocalDateTime
+import javax.inject.Inject
+
+class GoalServiceImpl @Inject constructor(
+    private val database: FirebaseFirestore,
+    private val experienceService: ExperienceService,
+    private val pointService: PointService,
+    private val userService: UserService
+) : GoalService {
+    override suspend fun create(
+        activity: SportActivity,
+        endDate: LocalDate,
+        email: String,
+        goal: Double,
+        monthly: Boolean,
+        startDate: LocalDate,
+        steps: Boolean
+    ) {
+        val reference = database
+            .collection(identifier)
+            .document()
+
+        val goalEntry = Goal(
+            activity = activity,
+            completed = false,
+            endDate = endDate,
+            email = email,
+            goal = goal,
+            monthly = monthly,
+            id = reference.id,
+            points = Points(),
+            progress = 0.0,
+            startDate = startDate,
+            steps = steps
+        )
+
+        goalEntry.points = goalEntry.points.copy(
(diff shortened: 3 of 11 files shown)


klaidoshka:
Wouldn't showing activity as a goal title be sufficient? By that I mean that we could drop the `weekly|monthly goal` part and it would still be informative enough. Since at the very top we can see at which tab/screen/page/part we are - either weekly or monthly. Where it is weekly, only weekly goals are shown within and it's the opposite with monthly.

app/src/main/java/com/pvp/app/api/GoalService.kt
```diff
@@ -0,0 +1,47 @@
+package com.pvp.app.api
+
+import com.pvp.app.model.Goal
+import com.pvp.app.model.SportActivity
+import kotlinx.coroutines.flow.Flow
+import java.time.LocalDate
+
+interface GoalService: DocumentsCollection {
```
klaidoshka, line 8: I think a space is missing before `:`

app/src/main/java/com/pvp/app/api/GoalService.kt
```diff
@@ -0,0 +1,47 @@
...
+        activity: SportActivity,
+        endDate: LocalDate,
+        email: String,
+        goal: Double,
+        monthly: Boolean,
+        startDate: LocalDate,
+        steps: Boolean
+    )
+
+    suspend fun claim(
```
klaidoshka, line 26: Could be single line. Missing comment.

app/src/main/java/com/pvp/app/api/PointService.kt
```diff
@@ -32,6 +33,16 @@ interface PointService {
         return tasks.sumOf { calculate(it) }
     }
 
+    /**
+     * Calculate the points of a goal. Points are determined by the goal type and the attributes of
+     * the goal. User status can also be taken into account, hence, [Goal.email] is required.
```
klaidoshka, line 38: _can be taken into account_ / _is required_, those are two opposites, which is it then? 😄
Also, if you call it required, then an error would be expected when email is not provided.

app/src/main/java/com/pvp/app/service/GoalServiceImpl.kt
```diff
@@ -0,0 +1,166 @@
...
+                    reference,
+                    JSON
+                        .encodeToJsonElement<Goal>(goalEntry)
+                        .toPrimitivesMap()
+                )
+            }
+            .await()
+    }
+
+    override suspend fun claim(
```
klaidoshka, line 76: Could be single line.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalCreation.kt
```diff
@@ -0,0 +1,383 @@
...
+                },
+                putBelow = true,
+                text = "${stateKilometers.value + (stateMeters.value / 1000.0)} (km) distance",
+                textAlign = TextAlign.End
+            )
+        },
+        dialogTitle = { Text("Editing distance") },
+        label = "Distance",
+        onConfirm = { onDistanceChange(stateKilometers.value + (stateMeters.value / 1000.0)) },
+        onDismiss = { },
```
klaidoshka, line 97: This parameter is most-likely optional, could drop it if not required.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalCreation.kt
```diff
@@ -0,0 +1,383 @@
...
+        value = "${steps.toInt()} steps"
+    )
+}
+
+@Composable
+private fun StepSelector(
+    isSelected: Boolean,
+    onClick: () -> Unit
+) {
+    Row(
```
klaidoshka, line 335: Using one of the `Tab` composables may would have been easier than this.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalScreen.kt
```diff
@@ -0,0 +1,338 @@
...
+            )
+
+            Spacer(modifier = Modifier.padding(8.dp))
+
+            ProgressBar(goal = goal)
+
+            Spacer(modifier = Modifier.padding(4.dp))
+
+            Row(
+                modifier = Modifier.padding(6.dp)
```
klaidoshka, line 169: Could be single line.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalScreen.kt
```diff
@@ -0,0 +1,338 @@
...
+}
+
+@Composable
+private fun GoalTypeSelector(
+    filter: GoalFilter,
+    isSelected: Boolean,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier
+) {
+    Box(contentAlignment = Alignment.Center,
```
klaidoshka, line 242: Put in new line; Parameters closing bracket will have to be shifted into new line too.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalScreen.kt
```diff
@@ -0,0 +1,338 @@
...
+                },
+                MaterialTheme.shapes.medium
+            )) {
+        Text(text = filter.displayName)
+    }
+}
+
+@Composable
+private fun DateChanger(
+    model: GoalViewModel = hiltViewModel(),
```
klaidoshka, line 259: Could be single line.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalViewModel.kt
```diff
@@ -0,0 +1,218 @@
...
+
+            val flowGoals = flowUser.flatMapLatest { user ->
+                user
+                    ?.let { goalService.get(user.email) }
+                    ?: flowOf(listOf())
+            }
+
+            flowUser
+                .combine(flowGoals) { user, goals ->
+                    user ?: return@combine _state.value
```
klaidoshka, line 55: Instead of such hacky return we could add `Flow#filterNotNull` onto the `userService.user` and then everything will be combined and collected only if user is not null. Then there will be no need to return empty goals list too.

Just reminding that `combine` waits until very first emit from any flow happens. So in case we know that user could possibly be null here for a long time, then thread would stay frozen, waiting. But I believe that here is not such case.

Also, could add `Dispatchers.IO` into `viewModelScope.launch(...)` so that everything happens off-main thread.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalViewModel.kt
```diff
@@ -0,0 +1,218 @@
...
+        steps: Boolean
+    ) {
+        val startDate = LocalDate.now()
+
+        val endDate = when (monthly) {
+            true -> startDate.plusMonths(1)
+            false -> startDate.plusDays(7)
+        }
+
+        viewModelScope.launch {
```
klaidoshka, line 97: Could be with IO dispatcher.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalViewModel.kt
```diff
@@ -0,0 +1,218 @@
...
+}
+
+data class GoalState(
+    val allGoals: List<Goal> = listOf(),
+    val currentGoals: List<Goal> = listOf(),
+    var monthly: Boolean = false,
+    var monthStartDate: LocalDate = LocalDate.now(),
+    var monthEndDate: LocalDate = LocalDate.now(),
+    val monthSteps: Long = 0,
+    val user: User? = null,
```
klaidoshka, line 215: Could become not-null if we know for certain that user is not going to be null in goal screens (passed authentication). For initial value pass `User()`.

Code could become simpler (at some places) with this change.

klaidoshka (approved):

app/src/main/java/com/pvp/app/ui/screen/goals/GoalScreen.kt
```diff
@@ -0,0 +1,338 @@
...
+                            item {
+                                Box(
+                                    contentAlignment = Alignment.Center,
+                                    modifier = Modifier.fillMaxWidth()
+                                ) {
+                                    Text(
+                                        fontStyle = FontStyle.Italic,
+                                        modifier = Modifier.padding(32.dp),
+                                        style = MaterialTheme.typography.bodyMedium,
+                                        text = "No goals have been set up yet."
```
klaidoshka, line 96: We may want to dispose the `.` at the end, since everywhere else (friend requests; tasks list; decorations) we are providing this text as title/label and not in sentence form (which would require punctuation).

app/src/main/java/com/pvp/app/api/PointService.kt
vytjar: The comment is just copied comment from `calculate` method for tasks, since it basically does the same. In that case, both need to be changed.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalViewModel.kt
vytjar: Since `CalendarWeeklyViewModel` and `CalendarMonthlyViewModel` both use the same return, will modify them too.

PR #99, opened 2024-04-29
PVP 246

vytjar:
Ensured that daily tasks contain at least 1 walking/wheelchair task.

- Added hasDisablity flag in the user model
- Disablity status is checked in the ActivityWorker, as it needs to be updated, because it could change.

### DIFF
--- a/app/src/main/java/com/pvp/app/service/TaskServiceImpl.kt
+++ b/app/src/main/java/com/pvp/app/service/TaskServiceImpl.kt
@@ -262,7 +262,7 @@ class TaskServiceImpl @Inject constructor(
         hasDisability: Boolean,
         userEmail: String
     ): List<SportTask> {
-        return getSportActivities(
+        return getDailySportActivities(
             count = count,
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/pvp/app/service/TaskServiceImpl.kt
```diff
@@ -482,6 +482,22 @@ class TaskServiceImpl @Inject constructor(
...
+        ): List<SportActivity> {
+            return if (hasDisability) {
+                listOf(SportActivity.Wheelchair)
+            } else {
+                SportActivity.entries
+                    .minus(SportActivity.Wheelchair)
+                    .minus(SportActivity.Other)
+                    .shuffled()
+                    .take(count - 1)
+                    .plus(SportActivity.Walking)
```
klaidoshka, line 497: Is it possible that we can have x2 walking daily tasks? Maybe we should not?

klaidoshka (approved):

PR #100, opened 2024-04-30
PVP-256 Adjust UI to have less dialogs, rework flows, add loading states

klaidoshka:
# Changes
- `Sidebar` is opened only by clicking the drawer icon-button. No more dragging.
- Routes are now registered as a `Root` (inner NavGraph) and as a `Node` (single route). This allows us to register inner graphs with routes that share the same view model/s in associated backstack.
- Changed navigation animation (when navigating between screens with navigation controller. this is not the case for `HorizontalPager` that allows us to access `Profile` screen)
- Friend preview was transferred to another screen.
- `FriendViewModel` code was reworked to use states instead of multiple variables.
- Some other `ViewModels` were adjusted to use flows in a more correct form (in my opinion), also coroutines are using IO dispatcher in most places now.
- Dialog to create task moved to bottom sheet.
- Dialog to edit task moved to bottom sheet.
- Added a few new `Local` variables that help to handle dynamic route options: icon, title.
  - Options are handled statically by Router, but in route's composable we can alter it further depending on our set conditions. Currently Friend and Profile screens are doing exactly that.
- Transferred some logic regarding tasks from `Layout` to `Calendar` screen.
- Added loading states to pages that load data: `profile`, `friends`, `friend`, `decorations` and using new `ProgressIndicatorWithinDialog` component that allows us to see changes being made under the dialog, while composable is working with currently-loading data.
  - `Calendar` screen is not using it, I think we may not need such logic there.
- Tasks preview form by default has `Daily` tasks group set instead of `General`.
- Tasks edit/create forms are using `Picker` for `Duration`, `ReminderTime` properties instead of `Slider`.
  - Also, Sport task creation form by default has `Walking` activity set and it is shown to the user. Before it wasn't shown to the user, but we defaulted to `Walking` activity either way, if users didn't set activity themselves.

### DIFF
--- a/app/src/main/java/com/pvp/app/ui/common/RouteUtil.kt
+++ b/app/src/main/java/com/pvp/app/ui/common/RouteUtil.kt
@@ -1,12 +1,7 @@
 package com.pvp.app.ui.common
 
-import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
-import androidx.compose.ui.graphics.vector.ImageVector
-import androidx.compose.ui.res.stringResource
-import androidx.compose.ui.unit.sp
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.ViewModel
 import androidx.navigation.NavBackStackEntry
@@ -29,24 +24,5 @@ object RouteUtil {
         }
 
         return hiltViewModel(parentEntry)
-    }
-
-    @Composable
-    fun RouteIcon(
-        imageVector: ImageVector,
-        resourceId: Int
-    ) {
-        androidx.compose.material3.Icon(
-            contentDescription = "${stringResource(resourceId)} route button icon",
-            imageVector = imageVector
-        )
-    }
-
-    @Composable
-    fun RouteTitle(title: String) {
-        Text(
-            style = MaterialTheme.typography.titleLarge.copy(fontSize = 24.sp),
-            text = title
-        )
     }
 }
--- a/app/src/main/java/com/pvp/app/ui/common/RouteUtil.kt
+++ b/app/src/main/java/com/pvp/app/ui/common/RouteUtil.kt
@@ -1,0 +1,52 @@
+package com.pvp.app.ui.common
+
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.sp
+import androidx.hilt.navigation.compose.hiltViewModel
+import androidx.lifecycle.ViewModel
+import androidx.navigation.NavBackStackEntry
+import androidx.navigation.NavController
+
+object RouteUtil {
+
+    /**
+     * If there is any routes in the backstack that had specified viewModel, it will be
+     * returned. Otherwise, it will return the viewModel of the current route.
+     */
+    @Composable
+    inline fun <reified T : ViewModel> NavBackStackEntry.hiltViewModel(
+        controller: NavController
+    ): T {
+        val navGraphRoute = destination.parent?.route ?: return hiltViewModel()
+
+        val parentEntry = remember(this) {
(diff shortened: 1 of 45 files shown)


app/src/main/java/com/pvp/app/ui/common/RouteUtil.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        val parentEntry = remember(this) {
+            controller.getBackStackEntry(navGraphRoute)
+        }
+
+        return hiltViewModel(parentEntry)
+    }
+
+    @Composable
+    fun RouteIcon(
```
klaidoshka, line 35: Not sure if this is a correct place for these components. Maybe they should be transferred to `RouteComponents.kt` or something similar...

app/src/main/java/com/pvp/app/ui/router/Router.kt
```diff
@@ -1,48 +1,123 @@
...
+}
+
+private fun NavGraphBuilder.composeRoute(
+    controller: NavHostController,
+    route: Route.Node,
+    routeModifier: Modifier
+) {
+    composable(route.path) { backstack ->
+        val settled = LocalHorizontalPagerSettled.current
+        var applierRequired by remember(settled) { mutableStateOf(settled) }
```
klaidoshka, line 104: Can't use it within `LaunchedEffect` block, because `LocalRouteOptionsApplier` is a composable function.

app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsCreate.kt
```diff
@@ -163,241 +161,210 @@ fun TaskCreateForm(
         }
     }
 
-    Surface(
```
klaidoshka, line 166: In whole file here and `TaskFormsEdit` - only removed redundant `Surface` wrapper, changed `when`, `if` into one `if` chain, moved `Slider` out, put `Picker` in (duration; reminderTime).

This creation form also had task type selector, so I transferred it into a different private composable to make it easier to read.

app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsEdit.kt
```diff
@@ -69,269 +70,240 @@ fun TaskEditDialog(
         else -> "Description"
     }
 
-    Dialog(onDismissRequest = onDialogClose) {
```
klaidoshka, line 72: Altered the same as with `TaskFormsCreate` file, except here are no task type selectors. So, only `when`, `if` changed to `if` chain and `Slider` swapped with `Picker` for `duration` and `reminderTime` properties.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendScreen.kt
```diff
@@ -0,0 +1,417 @@
...
+import com.pvp.app.ui.common.ProgressIndicatorWithinDialog
+import com.pvp.app.ui.common.RouteUtil.RouteTitle
+import com.pvp.app.ui.router.Route
+import com.pvp.app.ui.router.Routes
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+
+@Composable
```
klaidoshka, line 60: Everything copy-pasted from FriendsScreen. Except added `HandleState`, `RouteOptionsApplier`.

app/src/main/java/com/pvp/app/ui/router/Routes.kt
```diff
@@ -10,139 +10,232 @@ import androidx.compose.material.icons.outlined.Settings
...
+
+object Routes {
+
+    /**
+     * These routes are used when user state is set to **authenticated** and all required
+     * surveys are filled.
+     *
+     * Routes are used within [com.pvp.app.ui.screen.layout.LayoutScreenAuthenticated] layout.
+     */
+    val routesAuthenticated = listOf(
```
klaidoshka, line 71: Maybe the property names should have their `routes` prefix dropped?

app/src/main/java/com/pvp/app/ui/common/RouteUtil.kt
vytjar: RouteComponents would probably be a better name, yes

vytjar (approved):

PR #101, opened 2024-04-30
PVP-248 implement goal autocomplete

vytjar:
Implemented autocomplete for goals.

- Renamed TaskAutocompleteService and TaskAutocompleteWorker to AutocompleteService and AutocompleteWorker
- Implemented autocompletion logic in AutocompleteService
- Goal points are claimed upon autocompletion

### DIFF
--- a/app/src/main/java/com/pvp/app/service/AutocompleteService.kt
+++ b/app/src/main/java/com/pvp/app/service/AutocompleteService.kt
@@ -275,6 +275,10 @@ class AutocompleteService : Service() {
     private suspend fun updateGoals(goals: List<Goal>) {
         goals.forEach { goal ->
             goalService.update(goal)
+
+            if (goal.completed && goal.points.claimedAt != null) {
+                goalService.claim(goal)
+            }
         }
     }
 
--- a/app/src/main/java/com/pvp/app/service/AutocompleteService.kt
+++ b/app/src/main/java/com/pvp/app/service/AutocompleteService.kt
@@ -30,7 +30,7 @@ import java.time.ZoneId
 import javax.inject.Inject
 
 @AndroidEntryPoint
-class TaskAutocompleteService : Service() {
+class AutocompleteService : Service() {
 
     @Inject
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/pvp/app/service/AutocompleteService.kt
```diff
@@ -100,27 +105,112 @@ class TaskAutocompleteService : Service() {
...
+                false -> {
+                    goal.progress = getDistance(
+                        exercises,
+                        goal
+                    )
+                }
+            }
+
+            if (goal.progress >= goal.target) {
+                goal.completed = true
```
klaidoshka, line 142: I believe `goal.completed = goal.progress >= goal.target` would be simpler while giving the same result (since we are already working with not completed goals, so if we mark them false again, no problem). Once again, just my opinion. If you prefer this way, leave as it is, it is completely fine.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalScreen.kt
```diff
@@ -330,7 +330,7 @@ fun ProgressBar(goal: Goal) {
         Text(
             text = when (goal.steps) {
                 true -> "${goal.progress.toInt()} / ${goal.target.toInt()} steps"
-                false -> "${goal.progress} / ${goal.target} km"
+                false -> "${String.format("%.2f", goal.progress)} / ${goal.target} km"
```
klaidoshka, line 333: A shorter version, I believe, would be `"%.2f".format(goal.progress)`. Not required, just noting, maybe you didn't know and would prefer the shorter version, either is fine.

klaidoshka (approved):

app/src/main/java/com/pvp/app/ui/screen/goals/GoalScreen.kt
vytjar: Didn't know, thanks.

PR #102, opened 2024-05-01
PVP-239 Make EditableInfoItem smaller when it's empty

AurisTFG:
# Changes:
- Refactored/simplified Task Create/Edit Forms
  - Both forms now use the same method to avoid repeating code.
  - Introduced new methods like EditableTextItem, EditablePickerItem, EditableTimeItem, EditableDistanceItem, etc to reduce code repetition, improve readability and cleanness of the code by reducing the number of parameters and eliminating the need of separate temporary variables for editing.
  - Empty or null values now don't render a Text component, making the item boxes smaller.
  
# Other Fixes:
- Added a Spacer at the bottom of the Task Create/Edit forms to prevent overlap between phone navigation buttons and the app's buttons.
- Included rangeDuration and rangeMeters in Configuration for consistency across the entire configuration system.
- Ensured that configurations are utilized in all relevant places to maintain synchronization.
- Adjusted tooltip icon size to eliminate unnecessary whitespace usage.
- Changed DatePicker dialog to default to date picking mode instead of date input mode for quicker access.
- Corrected text for empty weekly activities box in the profile screen (changed from "No weekly activities selected" to "No weekly activities have been assigned yet").
- Updated mass and height editing in the profile screen to use pickers instead of the previous text input method, aligning with the consistency of the survey screen.
- Disabled clickability of cards in the profile screen (previously, they triggered a "clicked" animation, which is unnecessary as they shouldn't be clickable).

### DIFF
--- a/app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsCommon.kt
+++ b/app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsCommon.kt
@@ -1,190 +1,278 @@
 package com.pvp.app.ui.screen.calendar
 
+import androidx.compose.foundation.BorderStroke
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
-import androidx.compose.material3.DropdownMenuItem
-import androidx.compose.material3.ExperimentalMaterial3Api
-import androidx.compose.material3.ExposedDropdownMenuBox
-import androidx.compose.material3.ExposedDropdownMenuDefaults
-import androidx.compose.material3.Slider
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.Delete
+import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.Icon
+import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
-import androidx.compose.material3.TextField
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableDoubleStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.text.TextStyle
-import androidx.compose.ui.text.style.TextAlign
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsCommon.kt
```diff
@@ -1,190 +1,278 @@
...
-                    InfoTooltip(tooltipText = "This task is likely to be autocompleted")
-                }
-            }
+    var title by remember { mutableStateOf(task?.title ?: "") }
+    var description by remember {
+        mutableStateOf(
+            (task as? MealTask)?.recipe ?: task?.description ?: ""
+        )
+    }
+    var activity by remember {
```
klaidoshka, line 65: Every multi-line block should be separated by single empty line (above/under).

app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsCommon.kt
```diff
@@ -1,190 +1,278 @@
...
+                items = model.rangeDuration,
+                itemsLabels = "minutes",
+                onValueChange = { duration = it }
+            )
+        }
+
+        EditableTimeItem(
+            label = "Scheduled at",
+            value = dateTime.toLocalTime(),
+            valueDisplay = if ((taskClass == SportTask::class && activity.supportsDistanceMetrics)
```
klaidoshka, line 139: Split after every logical operator `&&`, `||`. Each condition should be in its own line. First condition should also be in its own line, not behind `if (`

app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsEdit.kt
```diff
@@ -2,310 +2,42 @@
...
-import java.time.Duration
-import java.time.LocalTime
-import java.time.format.DateTimeFormatter
 
 @Composable
 fun TaskEditSheet(
     task: Task,
-    onDialogClose: () -> Unit,
-    model: TaskViewModel = hiltViewModel()
+    onDialogClose: () -> Unit
```
klaidoshka, line 18: Oh, this could be renamed to `onSheetClose` or just `onClose`.

app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsEdit.kt
```diff
@@ -2,310 +2,42 @@
...
-            }
-        }
+        Spacer(modifier = Modifier.size(16.dp))
     }
 }
+
+@Composable
+fun TaskEditForm(
+    task: Task,
+    onDialogClose: () -> Unit
```
klaidoshka, line 37: Same as above: `onClose` (preferred in my opinion) or `onSheetClose`

klaidoshka (approved):

vytjar (approved):

PR #104, opened 2024-05-01
PVP-249 Applied reward popup design

vytjar:
Applied reward pop up design to RewardDialog and moved the composable to its own file.

Now it looks like this:

image

### DIFF
--- a/app/src/main/java/com/pvp/app/ui/screen/layout/RewardDialog.kt
+++ b/app/src/main/java/com/pvp/app/ui/screen/layout/RewardDialog.kt
@@ -9,13 +9,11 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.outlined.Star
 import androidx.compose.material.icons.outlined.Stars
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/pvp/app/ui/screen/layout/RewardDialog.kt
```diff
@@ -0,0 +1,160 @@
...
+    onClose: () -> Unit,
+    reward: Reward,
+    streak: Streak
+) {
+    if (!isOpen) {
+        return
+    }
+
+    Dialog(onDismissRequest = onClose) {
+        Surface(
```
klaidoshka, line 76: This looks redundant, can move those params into inner Column container. Probably Surface's `color` is backgroundColor, so inner column already overrides it with its modifier.

app/src/main/java/com/pvp/app/ui/screen/layout/RewardDialog.kt
```diff
@@ -0,0 +1,160 @@
...
+
+                reward.decoration?.let {
+                    DecorationCard(decoration = it)
+
+                    Spacer(modifier = Modifier.padding(8.dp))
+                }
+
+
+                Button(
+                    onClick = { onClose() }
```
klaidoshka, line 153: Could be one line.

klaidoshka (approved):

app/src/main/java/com/pvp/app/ui/screen/layout/RewardDialog.kt
vytjar: Surface is necessary to make the dialog have rounded corners. Color can be removed though.

AurisTFG (approved):

app/src/main/java/com/pvp/app/ui/screen/layout/RewardDialog.kt
klaidoshka: But I mean, Column can already have both, so Surface is just another redundant nesting layer, no?
```kotlin
modifier = Modifier
    .clip(MaterialTheme.shapes.medium) // instead of RoundedCornerShape(16.dp)
    .background(MaterialTheme.colorScheme.surface) // or the one Column already uses, since it overrides Surface's color 🤷
```

vytjar: Oh ffs I forgot you need to clip before setting the background, that's why it didn't work on the column before..

PR #113, opened 2024-05-07
PVP-261 ensured click shape matches components shape

vytjar:
Fixed some clickable box components so that the shape of the component and the click effect for that component match.


klaidoshka:
Screenshot_2024-05-08-10-54-44-331_com pvp app

When in dark theme, text of the button cannot be seen in the authentication screen. You could hard code it to black, since no matter the theme we are making authentication screen be _light_.

klaidoshka (approved):

PR #116, opened 2024-05-09
PVP-260 Fix sorting in friend list

AurisTFG:
# Changes: 
- Fixed sorting in friend list
- Updated gradle and all dependencies
- Refactored settings service in order to make it work with datastore versions above 1.1.0


app/src/main/java/com/pvp/app/api/SettingService.kt
```diff
@@ -26,4 +28,12 @@ interface SettingService {
         setting: Setting<T>,
         value: T?
     )
+
+    /**
+     * Remember the value of the setting, used in Composable functions.
+     */
+    @Composable
+    fun <T> remember(
+        setting: Setting<T>
```
klaidoshka, line 37: Could be single line.

app/src/main/java/com/pvp/app/service/SettingServiceImpl.kt
```diff
@@ -41,4 +48,36 @@ class SettingServiceImpl @Inject constructor(
 
         context.dataStoreSettings.edit { it[setting.key] = value }
     }
+
+    @Composable
+    override fun <T> remember(
```
klaidoshka, line 53: Could be single line. Also very weird that we have UI logic in business logic. If you can think of a way to make this work without `composable` being _here_ in the business layer, then it would be the preferred way. If not, let it be as it is.

app/src/main/java/com/pvp/app/service/SettingServiceImpl.kt
```diff
@@ -41,4 +48,36 @@ class SettingServiceImpl @Inject constructor(
...
+        }
+            .collectAsState(initial = setting.defaultValue)
+
+        return androidx.compose.runtime.remember {
+            object : MutableState<T> {
+                override var value: T
+                    get() = state.value
+                    set(value) {
+                        coroutineScope.launch {
+                            Log.e("SettingServiceImpl", "Setting value: $value")
```
klaidoshka, line 71: Why is this an error log? Also should be separated.

app/src/main/java/com/pvp/app/service/SettingServiceImpl.kt
```diff
@@ -41,4 +48,36 @@ class SettingServiceImpl @Inject constructor(
...
         context.dataStoreSettings.edit { it[setting.key] = value }
     }
+
+    @Composable
+    override fun <T> remember(
+        setting: Setting<T>
+    ): MutableState<T> {
+        val coroutineScope = rememberCoroutineScope()
+        val context = LocalContext.current
+        val state = androidx.compose.runtime.remember {
```
klaidoshka, line 58: Should be separated (because multi-line)

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsViewModel.kt
```diff
@@ -25,23 +24,11 @@ class SettingsViewModel @Inject constructor(
...
-    fun <T> merge(
-        setting: Setting<T>,
-        value: T
-    ) {
-        viewModelScope.launch(Dispatchers.IO) {
-            settingService.merge(setting, value)
-        }
+    @Composable
+    fun <T> rememberSetting(
+        setting: Setting<T>
```
klaidoshka, line 29: Could be single line.

klaidoshka (approved):

PR #117, opened 2024-05-14
PVP 271

vytjar:
- Ensured app does not crash if it gets a decoration from database which is of type that is not implemented in the current app version
- Added decoration type 'accessory'. Currently a red balloon of that type is created

image

### DIFF
--- a/app/src/main/java/com/pvp/app/model/Decoration.kt
+++ b/app/src/main/java/com/pvp/app/model/Decoration.kt
@@ -13,6 +13,7 @@ data class Decoration(
 )
 
 enum class Type {
+    AVATAR_ACCESSORY,
     AVATAR_BODY,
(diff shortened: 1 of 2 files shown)


klaidoshka:
I think we are using capitalized words in decoration names, let's be consistent. `Red balloon` -> `Red Balloon`, this should also be done to other decorations, if there are any.

DecorationServiceImpl has a sorting extension within companion object if I recall correctly. There you can provide in what order decoration should be drawn, now I believe accessory will be the last to be drawn. In case we add one more type in the future, I think it will be drawn in a random order (n'th or n-1'th). Maybe it should be explicitly defined that accessory is the last to be drawn for current types?

klaidoshka (approved):

AurisTFG (approved):

PR #118, opened 2024-05-14
PVP-265: Make sure that old daily tasks that weren't completed are removed

youreondrugs:
the old daily tasks are removed if not completed. if completed, they stay as archive and user cannot uncomplete them.


app/src/main/java/com/pvp/app/worker/DailyTaskWorker.kt
```diff
@@ -56,16 +59,14 @@ class DailyTaskWorker @AssistedInject constructor(
                         }
                         .first()
 
-                    val today = LocalDate.now()
+                    tasks.filter { it.date.isBefore(today) && !it.isCompleted }.forEach { task ->
```
klaidoshka, line 62: Should be split on each method call

app/src/main/java/com/pvp/app/ui/screen/calendar/TaskCards.kt
```diff
@@ -87,25 +88,27 @@ fun TaskCard(
...
-                        model.update(
-                            { task ->
-                                Task.copy(
-                                    task,
-                                    isCompleted = it
-                                )
-                            },
-                            task
-                        )
+                if (task.date == LocalDate.now()) {
```
klaidoshka, line 91: I believe you made it so that no task that is in the past/future can be checked as completed. Some of our functionalities are specifically made to support past/future completions. We should discuss how we want to handle this, because if we drop support for past/future completions, I believe you will also have to adjust the surrounding code (workers, task service, maybe something else).

klaidoshka (changes requested):

klaidoshka (approved):

PR #120, opened 2024-05-15
PVP-262

KhajiitLikeToSneak:
Modernized UI elements, by making adjustments and trying to keep the same design through every screen

- Brighter background
- Containers in the screen have darker background
- Containers are always with round shape (MaterialTheme.shapes.medium)
- Same horizontal padding


app/src/main/java/com/pvp/app/ui/common/Buttons.kt
```diff
@@ -60,7 +61,7 @@ fun Button(
 @Composable
 fun ButtonConfirm(
     border: BorderStroke? = null,
-    confirmationButtonContent: @Composable RowScope.() -> Unit = { Text("Proceed") },
+    confirmationButtonContent: @Composable RowScope.() -> Unit = { Text("Proceed", fontWeight = FontWeight.Bold) },
```
klaidoshka, line 64: I believe lambda body should be split into multiple lines, because Text has multiple parameters. There is `projectStyles.xml` under `./app`, try to import it to your IDE's Code Style Kotlin settings, `CTRL+ALT+L` would help you here.

app/src/main/java/com/pvp/app/ui/common/Buttons.kt
```diff
@@ -95,7 +96,7 @@ fun ButtonWithDialog(
     content: @Composable RowScope.() -> Unit = { Text("Open Dialog") },
     contentAlignment: Alignment = Alignment.TopStart,
     contentPadding: PaddingValues = ButtonDefaults.ContentPadding,
-    confirmButtonContent: @Composable RowScope.() -> Unit = { Text("Confirm") },
+    confirmButtonContent: @Composable RowScope.() -> Unit = { Text("Confirm", fontWeight = FontWeight.Bold) },
```
klaidoshka, line 99: Same as above

app/src/main/java/com/pvp/app/ui/screen/calendar/CalendarMonthlyScreen.kt
```diff
@@ -128,7 +133,11 @@ private fun ContentSwitch(
                     date
                 )
             } else {
-                TasksOfDay(tasks = tasks)
+                Column(
+                    modifier = Modifier.padding(top = 4.dp)
```
klaidoshka, line 137: Could be single line most likely, unless doesn't fit in the screen.

app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsCreate.kt
```diff
@@ -39,10 +38,10 @@ private fun ColumnScope.TaskTypeSelector(onSelect: (KClass<out Task>) -> Unit) {
     var selectedTabIndex by remember { mutableIntStateOf(0) }
 
     PrimaryTabRow(
-        containerColor = MaterialTheme.colorScheme.primaryContainer,
+        containerColor = MaterialTheme.colorScheme.secondaryContainer,
+        divider = {},
```
klaidoshka, line 42: I believe it can be dropped altogether, is optional.

app/src/main/java/com/pvp/app/ui/screen/calendar/TaskFormsCreate.kt
```diff
@@ -96,8 +97,6 @@ fun TaskCreateSheet(
 
             Spacer(modifier = Modifier.size(16.dp))
 
-            HorizontalDivider()
-
             Spacer(modifier = Modifier.size(8.dp))
```
klaidoshka, line 100: Spacers could be merged together into one and become 24.dp size or... Just the 8.dp spacer could be dropped, since we wanted a gap from the divider, which is not the case now.

app/src/main/java/com/pvp/app/ui/screen/calendar/TasksOfDay.kt
```diff
@@ -107,26 +96,42 @@ private fun TaskTypeFilter(
...
-                onClick = { onClick(filterNew) },
-                modifier = Modifier
-                    .weight(1f)
-                    .fillMaxWidth()
-                    .height(40.dp)
-            )
+
+        PrimaryTabRow(
+            containerColor = MaterialTheme.colorScheme.secondaryContainer,
+            divider = {},
```
klaidoshka, line 115: Could be dropped.

app/src/main/java/com/pvp/app/ui/screen/decoration/Decoration.kt
```diff
@@ -44,13 +45,18 @@ private fun DecorationCard(
...
             .background(
-                color = if (holder.applied) {
-                    MaterialTheme.colorScheme.surfaceContainer.darken(0.15f)
-                } else {
-                    MaterialTheme.colorScheme.surfaceContainer
-                },
+                color = MaterialTheme.colorScheme.surfaceContainer.darken(0.2f),
                 shape = MaterialTheme.shapes.medium
             )
+            .let {
```
klaidoshka, line 51: `Modifier.then` exists for this purpose, which appends specified Modifier onto current Modifier instance. But this works too, may be less code and similar readability.

app/src/main/java/com/pvp/app/ui/screen/decoration/DecorationScreen.kt
```diff
@@ -118,16 +117,13 @@ fun DecorationScreen(modifier: Modifier) {
...
-            .background(MaterialTheme.colorScheme.background)
+            .padding(16.dp)
     ) {
-        Spacer(modifier = Modifier.size(16.dp))
-
         PrimaryTabRow(
-            containerColor = MaterialTheme.colorScheme.primaryContainer,
-            divider = { VerticalDivider() },
+            containerColor = MaterialTheme.colorScheme.secondaryContainer,
+            divider = {},
```
klaidoshka, line 124: Can be dropped.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendScreen.kt
```diff
@@ -171,75 +179,75 @@ private fun Content(
...
-                        )
+                                .background(
+                                    MaterialTheme.colorScheme.surfaceContainerHighest.lighten(
+                                        0.08f
+                                    )
+                                )
+                        ) {
+                            Image(
+                                bitmap = mutualFriend.avatar,
+                                contentDescription = "Friend avatar",
```
klaidoshka, line 242: `"${mutualFriend.user.username} avatar"` may be more correct.

app/src/main/java/com/pvp/app/ui/screen/friends/FriendsScreen.kt
```diff
@@ -195,13 +203,21 @@ fun FriendsScreen(
                         val selectedTab = remember { mutableIntStateOf(0) }
 
                         Column {
-                            TabRow(selectedTabIndex = selectedTab.intValue) {
+                            PrimaryTabRow(
+                                containerColor = MaterialTheme.colorScheme.secondaryContainer,
+                                divider = {},
```
klaidoshka, line 208: Can be dropped.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalCreation.kt
```diff
@@ -328,56 +337,41 @@ fun StepPicker(
...
-    Row(
-        horizontalArrangement = Arrangement.SpaceBetween,
+    val selectedTabIndex = when (selectedStepType) {
+        StepType.Steps -> 0
+        StepType.Distance -> 1
+    }
+
+    PrimaryTabRow(
+        containerColor = MaterialTheme.colorScheme.secondaryContainer,
+        divider = {},
```
klaidoshka, line 351: Can be dropped.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalCreation.kt
```diff
@@ -328,56 +337,41 @@ fun StepPicker(
...
-                .height(40.dp)
-                .clickable { onClick() }
-                .background(
-                    if (!isSelected) {
-                        MaterialTheme.colorScheme.secondaryContainer
-                    } else {
-                        Color.Transparent
-                    },
-                    MaterialTheme.shapes.medium
+        StepType.values().forEach { stepType ->
```
klaidoshka, line 357: `entries` instead of `values()` or else split.
`entries` returns a constant immutable list, while `values()` returns a new array each time.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalCreation.kt
```diff
@@ -328,56 +337,41 @@ fun StepPicker(
...
+                    fontWeight = if (stepType == selectedStepType) FontWeight.Bold else FontWeight.Normal
                 )
-        ) {
-            Text(text = "Distance")
+            }
         }
     }
+}
+
+enum class StepType(val displayName: String) {
```
klaidoshka, line 374: `DistanceType` probably would be the better name choice. Should rename associated parameters/variables above too.

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -284,24 +282,19 @@ private fun SettingApplicationTheme(
...
-        modifier = Modifier
-            .padding(
-                horizontal = 8.dp,
-                vertical = 4.dp
-            )
-            .fillMaxWidth(),
-        onClick = {
+    SettingCard(
+        title = "Health Connect",
+        description = "By enabling certain Health Connect permissions you are allowing us to provide you more great features. You can disable those permissions anytime by coming here.",
```
klaidoshka, line 287: Could this be split into few parts so that it wouldn't go out of the screen? Press `Enter` in certain spots, IDE will automatically split and join strings for you.

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -499,27 +485,39 @@ fun <T> SettingCard(
                     .align(Alignment.End)
                     .clickable {
                         if (isEnabled) {
-                            dialogOpen = true
+                            if (!isPermissions) {
```
klaidoshka, line 488: Maybe rename from `isPermissions` to `onEditShowDialog` or something similar, more abstract and there will be no need to rename it in the future, if another setting appears that requires the dialog to be not opened and is not associated with permissions anyhow. Or just make `editContent` nullable (wrap into `()?`) and check if it is null, if it is, means we have nothing to show in the dialog, so we don't.

app/src/main/java/com/pvp/app/ui/screen/statistics/StatisticsScreen.kt
```diff
@@ -115,11 +117,10 @@ fun StatisticsScreen(
         }
 
         PrimaryTabRow(
-            containerColor = MaterialTheme.colorScheme.primaryContainer,
+            containerColor = MaterialTheme.colorScheme.secondaryContainer,
+            divider = {},
```
klaidoshka, line 121: Can be dropped.

vytjar (approved):

PR #121, opened 2024-05-15
PVP-269 Add meal (with recipes) form body to the task creation form

klaidoshka:
# Changes

- Reworked common task creation form to use states
- Transferred task form's inputs to their own file
- Added more ingredients that most people are allergic to and also ingredients that are used in most meals to make
- Added new form type: Set Meals
  - User can select our own meal recipe by selecting it in a scrollable row.
  - Selected recipe is always shown the first in the recipes row, with a border and check mark
  - Recipes row can be filtered via a search input below
    - ❗ Selected meal is not shown here if no meals are found. Should this be adjusted?
- Meal recipe cards contain two tooltips
  - Breakdown tooltip: Shows meal name, calories, servings and ingredients breakdown
  - Ingredients to avoid tooltip: Shows ingredients that are contained in the recipe, but user chose to avoid them.
- Meal recipe card shows nutrition of the meal by % (carbohydrates/fat/protein)
- After selecting the meal, form is expanded: ingredients, recipe and meal title are added (non-editable fields)
- Created TabSelector common component, since it became used in way too many places
- Adjusted FoldableContent and some other components to meet my needs in this task
- Cleaned-up forgotten redundant logging
- Fixed bugs in Calendar view regarding task model changes:
  - Now general, meal tasks are counted/handled as expected
- Fixed a bug in profile where username is not being updated for the Delete button, so user sees `""` as required entry.
- Other mostly minor changes

# Edit

- Now showing selected meal even when search query returns nothing


app/src/main/java/com/pvp/app/ui/screen/decoration/Decoration.kt
```diff
@@ -140,7 +141,14 @@ fun DecorationCards(
                     }
                 }
             },
-            header = type.toString()
+            header = type.toString(),
+            modifier = Modifier
```
klaidoshka, line 145: Modifier became providable in the parameters for this component so that other places could adjust required style as wanted, it wouldn't be hard-coded.

vytjar (changes requested):
- There seem to be some problems with text displayed on food image readability in dark mode. (see Discord)
- When user creates a custom meal task, it is shown under general. Recipe field should be made mandatory.

vytjar (approved):

PR #125, opened 2024-05-16
#266 - Apply completed goal design

youreondrugs:
image


app/src/main/java/com/pvp/app/ui/screen/goals/GoalScreen.kt
```diff
@@ -108,6 +110,28 @@ fun GoalScreen(
...
+
+                        item {
+                            Spacer(modifier = Modifier.padding(30.dp))
+
+                            Text(
+                                style = MaterialTheme.typography.titleLarge,
+                                modifier = Modifier.fillMaxWidth(),
+                                textAlign = TextAlign.Center,
+                                text = "Completed:"
+                            )
```
klaidoshka, line 122: If we drop `:` and align text to the left, I think it will look better, just like with settings screen.
Possibly less space around the title will be needed too, try to make it similar to settings screen.

app/src/main/java/com/pvp/app/ui/screen/goals/GoalScreen.kt
```diff
@@ -210,6 +234,87 @@ fun GoalCard(
...
+            }
+
+            if (goal.steps) {
+                Spacer(modifier = Modifier.padding(2.dp))
+
+                Text(
+                    style = MaterialTheme.typography.bodyMedium,
+                    modifier = Modifier.padding(start = 6.dp),
+                    textAlign = TextAlign.Left,
+                    text = "Your average pace was " + when (goal.monthly) {
```
klaidoshka, line 304: I believe this is not really giving us average pace. We, most likely, want to show average pace in distance or steps **per day**. So if this was monthly goal, we divide sum by 30 or 31, or 28 (depends, unless we hard-coded it to 30 days when creating the goal). If it was weekly, we divide sum by 7.

klaidoshka (changes requested):

klaidoshka (approved):

PR #128, opened 2024-05-28
PVP 290 Changed toasts to snackbars

vytjar:
- Modified Composables that used toasts to be made of scaffolds that allow showing of snackbars
- Created CenteredSnackbarHost that centers the text which looks nicer

This is how they look now:

image

### DIFF
--- a/app/src/main/java/com/pvp/app/ui/screen/drawer/DrawerScreen.kt
+++ b/app/src/main/java/com/pvp/app/ui/screen/drawer/DrawerScreen.kt
@@ -1,5 +1,6 @@
 package com.pvp.app.ui.screen.drawer
 
+import android.annotation.SuppressLint
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
@@ -17,22 +18,25 @@ import androidx.compose.foundation.lazy.items
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.ModalDrawerSheet
+import androidx.compose.material3.Scaffold
+import androidx.compose.material3.SnackbarHostState
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.RectangleShape
-import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import androidx.hilt.navigation.compose.hiltViewModel
 import com.pvp.app.R
 import com.pvp.app.ui.common.ButtonConfirm
+import com.pvp.app.ui.common.CenteredSnackbarHost
 import com.pvp.app.ui.common.RouteTitle
-import com.pvp.app.ui.common.showToast
 import com.pvp.app.ui.router.Route
+import kotlinx.coroutines.launch
 
 @Composable
 private fun Body(
@@ -149,6 +153,7 @@ private fun Header(modifier: Modifier = Modifier) {
     }
 }
 
+@SuppressLint("UnusedMaterial3ScaffoldPaddingParameter")
 @Composable
 fun DrawerScreen(
     onClick: Route.() -> Unit,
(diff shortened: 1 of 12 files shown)


klaidoshka:
Cannot test it myself, so I will just give some comments and for now accept it, as I believe it works to some extent.

From user perspective.
- How is it being closed? By swiping left/right/up/down? Or some other specific way?
- Could we add a close button to the very row end for aesthetics and intuitive close function?
  - And leave left/right swipe to close it too, if it already exists. If not, would be nice to have for experienced users.
- Maybe our logo would be nice to the very row start?

From code perspective.
- Couldn't we just add scaffold to the LayoutBootstrap and propagate `showScaffold` function to the children components via `CompositionLocalProvider`? This would save us from nesting each screen into additional scaffold container, which is only used for snackbar, where one singular snackbar is sufficient (I believe) and there is no need to have different ones for each screen.

vytjar:
> * How is it being closed? By swiping left/right/up/down? Or some other specific way?
> * Could we add a close button to the very row end for aesthetics and intuitive close function?

It closes itself, just like a toast.

vytjar:
> * Couldn't we just add scaffold to the LayoutBootstrap and propagate `showScaffold` function to the children components via `CompositionLocalProvider`? This would save us from nesting each screen into additional scaffold container, which is only used for snackbar, where one singular snackbar is sufficient (I believe) and there is no need to have different ones for each screen.

Thanks, didn't know about that. Seems to work.

app/src/main/java/com/pvp/app/ui/screen/drawer/DrawerScreen.kt
```diff
@@ -149,13 +149,16 @@ private fun Header(modifier: Modifier = Modifier) {
     }
 }
 
+@SuppressLint("UnusedMaterial3ScaffoldPaddingParameter")
```
klaidoshka, line 152: +1

app/src/main/java/com/pvp/app/ui/screen/profile/ProfileScreen.kt
```diff
@@ -72,13 +71,14 @@ private val INGREDIENTS = Ingredient.entries.map { it.title }
 @Composable
 private fun AccountDeleteButton(
     viewModel: ProfileViewModel = hiltViewModel(),
-    state: ProfileState
+    state: ProfileState,
```
klaidoshka, line 74: +1

app/src/main/java/com/pvp/app/ui/screen/profile/ProfileScreen.kt
```diff
@@ -268,7 +269,9 @@ fun ProfileScreen(
     ) {
         Points(points = state.user.points)
 
-        Column(modifier = Modifier.fillMaxWidth()) {
+        Column(
+            modifier = Modifier.fillMaxWidth()
```
klaidoshka, line 273: +1 (unless doesn't fit)

app/src/main/java/com/pvp/app/ui/screen/profile/ProfileScreen.kt
```diff
@@ -255,6 +255,7 @@ fun ProfileScreen(
 ) {
     val state by viewModel.state.collectAsStateWithLifecycle()
 
+
```
klaidoshka, line 258: +1

app/src/main/java/com/pvp/app/ui/screen/profile/ProfileScreen.kt
```diff
@@ -291,7 +294,7 @@ fun ProfileScreen(
                 onUpdateIngredients = { viewModel.update { u -> u.ingredients = it } },
                 onUpdateHeight = { viewModel.update { u -> u.height = it } },
                 onUpdateMass = { viewModel.update { u -> u.mass = it } },
-                state = state
+                state = state,
```
klaidoshka, line 297: +1

app/src/main/java/com/pvp/app/ui/screen/profile/ProfileScreen.kt
```diff
@@ -306,15 +309,14 @@ private fun Properties(
     onUpdateHeight: (Int) -> Unit,
     onUpdateIngredients: (List<Ingredient>) -> Unit,
     onUpdateMass: (Int) -> Unit,
-    state: ProfileState
+    state: ProfileState,
```
klaidoshka, line 312: +1

app/src/main/java/com/pvp/app/ui/screen/settings/SettingsScreen.kt
```diff
@@ -255,13 +255,16 @@ private fun SettingHealthConnectPermissions(context: Context) {
 }
 
 @Composable
-private fun GoogleCalendarSynchronizer(model: SettingsViewModel = hiltViewModel()) {
+private fun GoogleCalendarSynchronizer(
+    model: SettingsViewModel = hiltViewModel(),
```
klaidoshka, line 259: +1

AurisTFG (approved):

PR #129, opened 2024-05-28
PVP-283 Ensured weekly activities contain activities user participates in

vytjar:
The most frequent activity that the user participates in is always added to the weekly activities generated.

### DIFF
--- a/app/src/main/java/com/pvp/app/api/ExerciseService.kt
+++ b/app/src/main/java/com/pvp/app/api/ExerciseService.kt
@@ -26,4 +26,10 @@ interface ExerciseService {
      * @return a list of sport activities that the user does not participate in frequently
      */
     suspend fun getInfrequentActivities(maxOccurrence: Int = 1): List<SportActivity>
+
+    /**
+     * Reads user activity data from HealthConnect and finds the most frequently done activity
+     * @return the most frequently done activity
+     */
+    suspend fun getMostFrequentActivity(): SportActivity
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/pvp/app/api/ExerciseService.kt
```diff
@@ -26,4 +26,10 @@ interface ExerciseService {
      * @return a list of sport activities that the user does not participate in frequently
      */
     suspend fun getInfrequentActivities(maxOccurrence: Int = 1): List<SportActivity>
+
+    /**
+     * Reads user activity data from HealthConnect and finds the most frequently done activity
+     * @return the most frequently done activity
+     */
+    suspend fun getMostFrequentActivity(): SportActivity
```
klaidoshka, line 34: If there is none, i.e. user has a fresh phone/google account? What happens? Maybe we should fallback to `Walking` activity?

app/src/main/java/com/pvp/app/service/ExerciseServiceImpl.kt
```diff
@@ -120,4 +120,10 @@ class ExerciseServiceImpl @Inject constructor(
                     activity != SportActivity.Other
         }
     }
+
+    override suspend fun getMostFrequentActivity(): SportActivity {
```
klaidoshka, line 124: Ah, that is being done, what I mentioned above... Comment could be improved in the interface to let user know this 😁

klaidoshka (approved):

app/src/main/java/com/pvp/app/api/ExerciseService.kt
vytjar: Yes, that's exactly what happens. Will expand the comment to include that.

PR #134, opened 2024-05-30
PVP-276: Add notifications for weekly meals

youreondrugs:


app/src/main/java/com/pvp/app/ui/screen/goals/GoalScreen.kt
```diff
@@ -238,9 +238,10 @@ fun GoalCard(
                     style = MaterialTheme.typography.bodyMedium,
                     modifier = Modifier.padding(start = 6.dp),
                     textAlign = TextAlign.Left,
-                    text = "You're average " + when (goal.monthly) {
-                        true -> "monthly steps: %d".format((monthSteps / 30))
-                        false -> "weekly steps: %d".format((monthSteps / 30 * 7))
+                    text = if (goal.monthly) {
```
klaidoshka, line 241: Not _only here_, but the whole screen must be localized. Lines 277, 297 included.

app/src/main/java/com/pvp/app/worker/MealPlanWorker.kt
```diff
@@ -26,6 +35,10 @@ class MealPlanWorker @AssistedInject constructor(
 ) {
 
     override suspend fun doWork(): Result {
+        val meals = mealService.get().first()
```
klaidoshka, line 38: What are you doing with these meals next?

app/src/main/java/com/pvp/app/worker/MealPlanWorker.kt
```diff
@@ -26,6 +35,10 @@ class MealPlanWorker @AssistedInject constructor(
...
+
+        postActivityNotification()
```
klaidoshka, line 40: If user doesn't exist (Line 44), we are not generating meals, but notification is already sent that meals were generated. Move this line downwards to make sure that notification is sent **ONLY IF** meals were actually generated.

app/src/main/java/com/pvp/app/worker/MealPlanWorker.kt
```diff
@@ -26,10 +35,13 @@ class MealPlanWorker @AssistedInject constructor(
 ) {
 
     override suspend fun doWork(): Result {
+
```
klaidoshka, line 38: Redundant empty line

app/src/main/java/com/pvp/app/worker/MealPlanWorker.kt
```diff
@@ -26,10 +35,13 @@ class MealPlanWorker @AssistedInject constructor(
...
         val user = userService.user.firstOrNull()
 
         user ?: return Result.retry()
 
+        postActivityNotification()
```
klaidoshka, line 43: Line 53 checks if meal plan has been already generated this week, which means, checks if user already got such notification for generated plan.
```kotlin
        if (user.lastMealPlanGeneratedWeek >= week) {
            return Result.success()
        }
```

Currently user will get notification on every application start-up.

Move notification posting below that condition check, best would be to line 63 and separate with empty line from return statement as so:
```kotlin
        userService.merge(user)

        postActivityNotification()

        return Result.success()
```

app/src/main/res/values-lt/strings.xml
```diff
@@ -146,6 +146,9 @@
     <string name="global_level">Lygis</string>
     <string name="goals_card_average_monthly_steps">Vidutinis nueitų žingsnių kiekis per mėnesį %1$s</string>
     <string name="goals_card_average_weekly_steps">Vidutinis nueitų žingsnių kiekis per savaitę %1$s</string>
+    <string name="goals_card_completed">Siekis %1$s įvykdytas!</string>
+    <string name="goals_card_completed_description_km">Jūsų siekis buvo: %1$.2f km</string>
+    <string name="goals_card_completed_description_steps">Your goal was: %1$.2f žingsniai</string>
```
klaidoshka, line 151: `Your goal was:` -> `Jūsų siekis buvo:`
I would also suggest dropping those `:` in both locales, would look cleaner.

klaidoshka (changes requested):

klaidoshka (approved):

